PR #43, opened 2022-03-15
Update statements.py

Fohlen:
Currently you fetch 8989 :)


pudo:
Aargh, good catch. Thanks for the fix!

PR #315, opened 2023-08-15
support since parameter for incremental scans

### LINKED ISSUE #312: Implement incremental scans
cc @everplays - because you mentioned it.

There's two tiers to this:

1. Implement query parameters for the `/search` and `/match` APIs that allow a `since=` search. They would probably just filter on `since > entity.last_changed`. This gives you the ability to do incremental queries with the state managed externally.
2. Implement some sort of state in the API itself. In this scenario, a client would submit a `/match`-style entity and a callback mechanism (e.g. a webhook) and the API would then store this query, execute it at a set interval and notify the client whenever a new result is available. This could also be implemented a REST primitive, a la `POST /alerts` and `GET /alerts?since=xxx`.

I'm curious: which of these did you have in mind?

everplays:
fixes #312.

@pudo is this what you had in mind?

### DIFF
--- a/yente/routers/match.py
+++ b/yente/routers/match.py
@@ -63,6 +63,11 @@ async def match(
         settings.MATCH_FUZZY,
         title="Use slow matching for candidate generation, does not affect scores",
     ),
+    since: str = Query(
+        None,
+        pattern="^[0-9]{4}-[0-9]{2}-[0-9]{2}$",
+        title=f"Match against records that were updated since the given date",
+    )
 ) -> EntityMatchResponse:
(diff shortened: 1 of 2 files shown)


pudo:
Perfect! Do we know what this does with `last_change=null`? This doesn't happen with OpenSanctions data but people can load other stuff, too. 

p.s. I'm out on a trip for the rest of the week. Would it be OK if I merged this on Monday (and got more involved in the OpenSearch issue)?

everplays:
Good question. I'll figure it out and update it here. In the meantime, enjoy your trip :-)

PR #332, opened 2023-09-18
Add schema facet and option to specify which facets are included in the response

jbothma:
Because filter fields are queried differently from schema.

An alternative might be to have a schemata filter, where you can filter by one or more explicit schemata, rather than the current schema parameter which does the matchable descendents cleverness.


jbothma:
I wanted to suggest something like that considering our growing list of
datasets

PR #425, opened 2024-04-09
Use proxy in async client based on env var

### LINKED ISSUE #419: Allow httpx to use proxies
Dear team,

Related to https://github.com/opensanctions/yente/issues/84, since the switch to httpx for data retrieval we cannot use our proxy anymore and therefore cannot update Yente. Could you please add the option to specify a proxy, as in: https://www.python-httpx.org/advanced/proxies/ ? 

for example
`with httpx.Client(proxy=YENTE_PROXY) as client:
    ...`
where YENTE_PROXY is an env var. 

Thank you very much!

RMHogervorst:
Attempt to set proxy based on proxy env vars.
I figured I give this a try. 
I did run black. but pytest didn't work for me, probably some async specific thing. 
Fix #419

### DIFF
--- a/tests/unit/test_data.py
+++ b/tests/unit/test_data.py
@@ -3,8 +3,7 @@ from pathlib import Path
 
 from yente.data import get_catalog
 from yente.data.loader import load_json_lines
-from yente.data.util import get_url_local_path
-from yente.data.util import phonetic_names
+from yente.data.util import get_url_local_path, phonetic_names, _proxy_env
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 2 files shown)


RMHogervorst:
So this is rather ugly but my thinking was:
- people will most often set HTTP_PROXY
- and only sometimes ALL_PROXY

This does not deal with cases when you want to set a different http and https proxy (though I do not know if that happens)

SimonThordal:
Thanks a lot for submitting this, @RMHogervorst! We just added the possibility of explicitly setting `YENTE_HTTP_PROXY`, so I hope that will solve this for you? I'll close this PR for now, feel free to open it or comment in #419 if you're still seeing issues.

PR #428, opened 2024-04-15
feat: proxy requests if requested

### LINKED ISSUE #419: Allow httpx to use proxies

SimonThordal:
Closes #419

### DIFF
--- a/yente/data/util.py
+++ b/yente/data/util.py
@@ -12,6 +12,7 @@ from rigour.text.scripts import is_modern_alphabet
 from rigour.text.distance import levenshtein
 from fingerprints import remove_types, clean_name_light
 from nomenklatura.util import fingerprint_name, names_word_list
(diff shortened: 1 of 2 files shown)


SimonThordal:
> Looks good. My biggest concern is that we use not only httpx but also aiohttp somewhere, but I cannot find evidence of that right now with a simple search of the codebase.

Can't find it either, so let's merge it in :)

PR #429, opened 2024-04-15
feat: add query parameter for allowing datasets

### LINKED ISSUE #427: Support for `dataset` allow-listing in /search and /match APIs
We want to introduce a new query argument that will add support for `/match/default?dataset=eu_fsf&dataset=us_ofac_sdn` to let users pick the lists they want to screen in, rather than the having to define custom manifests or using `exclude_dataset`.

SimonThordal:
Closes #427

### DIFF
--- a/tests/unit/test_match.py
+++ b/tests/unit/test_match.py
@@ -102,6 +102,44 @@ def test_match_exclude_dataset():
     assert len(res["results"]) == 0, res
 
 
+def test_match_include_dataset():
+    # When querying Putin
+    query = {"queries": {"vv": EXAMPLE}}
+    # Using only datasets that do not include Putin
+    params = {
+        "algorithm": "name-based",
+        "include_dataset": ["ae_local_terrorists", "mx_governors"],
+    }
+    resp = client.post("/match/default", json=query, params=params)
+    # We should get a succesful response
+    assert resp.status_code == 200, resp.text
+    data = resp.json()
+    res = data["responses"]["vv"]
+    # And we should get no matches
+    assert len(res["results"]) == 0, res
+    # When using a dataset that includes Putin
+    params = {
+        "algorithm": "name-based",
+        "include_dataset": ["eu_fsf", "ae_local_terrorists"],
+    }
+    resp = client.post("/match/default", json=query, params=params)
+    data = resp.json()
+    res = data["responses"]["vv"]
+    # And we should get matches
+    assert len(res["results"]) > 0, res
+    # When we exclude the eu_fsf dataset
+    params = {
+        "algorithm": "name-based",
+        "include_dataset": ["eu_fsf", "mx_governors", "ae_local_terrorists"],
+        "exclude_dataset": "eu_fsf",
+    }
+    # We should get no matches
+    resp = client.post("/match/default", json=query, params=params)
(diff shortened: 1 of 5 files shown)


tests/unit/test_match.py
```diff
@@ -102,6 +102,44 @@ def test_match_exclude_dataset():
     assert len(res["results"]) == 0, res
 
 
+def test_match_include_dataset():
+    # When querying Putin
+    query = {"queries": {"vv": EXAMPLE}}
+    # Using only datasets that do not include Putin
+    params = {
+        "algorithm": "name-based",
+        "allow_datasets": ["ae_local_terrorists", "mx_governors"],
```
jbothma, line 111: A small improvement might be to use datasets that occur in the test data and a query which occurs in both, but I don't think this test is invalid because of this choice.

jbothma (approved):
Nice!

pudo:
Nitpick: we have `exclude_dataset` and now we're adding `allow_datasets` (plural). What if we tried to make these matching, e.g. `include_dataset` for the new one?

SimonThordal:
> Nitpick: we have exclude_dataset and now we're adding allow_datasets (plural). What if we tried to make these matching, e.g. include_dataset for the new one?

That was my first choice actually, but I changed it since the path parameter already is called `dataset` and I thought it made it clearer that we were further refining that first dataset / collection. I'm good with either though. As for plural it's for making it clear that it's a list and not a single string.

tests/unit/test_match.py
SimonThordal: Do you mean using two datasets where Putin occurs in both of them?

pudo:
Both of them are lists - only the `dataset` is part of the path, right?

tests/unit/test_match.py
pudo: In the unit tests for yente, only eu_fsf is actually present. So neither of these datasets are in the index.

SimonThordal: Gotcha. In that case I think all the tests are valid, though a bit wonky.

SimonThordal:
> Both of them are lists - only the dataset is part of the path, right?
True, that was part of my confusion at first. From the signature of `/match` it sounded as though I should specify a single dataset as the path parameter `dataset` and then I could pass a single dataset as `exclude_dataset`. 

It becomes clear once you read through the code of course, but I think something like `{collection}/match?exclude_datasets=[]&include_datasets=[]` would have been easier to understand at first glance. 
I'll switch back to `include_dataset` for now though and then we can always see whether we want to change this later :)

SimonThordal:
Just saw that we want to add this to /search as well, so I'll include that here.

tests/unit/test_match.py
jbothma: Sorry for the slow response. Yeah - it looks like the fixtures have fsf and parteispenden, and queries which have a match from both datasets would just be a bit more technically correct, since we can only expect matches from datasets in the index.

Something we're not testing for yet, is the behaviour for a single entity that was merged and occurs in both datasets. If only one of those datasets are allowed by query param, is the entity returned? And if any of its datasets are excluded by query param (but not all), is it excluded?

PR #447, opened 2024-05-16
Expose index from catalog

SimonThordal:
Updates to `/catalog` based off of Lunars request here:

https://opensanctions.slack.com/archives/C02L061S7KK/p1715354587822739

### DIFF
--- a/yente/debug.py
+++ b/yente/debug.py
@@ -1,0 +1,16 @@
+from uvicorn import run
+
+from yente import settings
+from yente.app import create_app
+
+app = create_app()
+
+if __name__ == "__main__":
+    run(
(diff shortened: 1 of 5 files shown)


SimonThordal:
@pudo you mentioned the dataset itself having a `self.rest` parameter that could define how we serialize the dataset. Did you mean so they could customize what was available through `/catalog`? In the end I went a simpler way since it sounded like they really wanted to be able to do codegen off of the API and not have to work directly with the json.

yente/debug.py
SimonThordal, line 1: This is just here to easily let me start Yente with a debugger.

PR #459, opened 2024-06-05
Use deltas

SimonThordal:


yente/search/indexer.py
```diff
@@ -202,3 +213,99 @@ async def update_in_thread() -> None:
...
+            if current_version == target_version:
+                log.info(
+                    "Dataset is up to date.",
+                    dataset=dataset.name,
+                    current_version=current_version,
+                )
+                continue
+            clone = await index.clone(target_version)
+            # Get the next version.
+            while next_version := await get_next_version(dataset, current_version):
```
SimonThordal, line 301: Might want to add some cycle detection here.

yente/search/indexer.py
```diff
@@ -202,3 +213,99 @@ async def update_in_thread() -> None:
...
+        return
+    next_version = available_versions[ix + 1]
+    return next_version
+
+
+async def delta_update_index(force: bool = True):
+    # Get the catalog of datasets
+    catalog = await get_catalog()
+    log.info("Index update check")
+    provider = await ESSearchProvider.create()
```
SimonThordal, line 282: This should be configurable.

yente/search/indexer.py
```diff
@@ -202,3 +213,99 @@ async def update_in_thread() -> None:
...
+    pass
+
+
+async def get_deltas_from_version(
+    version: str, dataset: Dataset
+) -> AsyncGenerator[str, None]:
+    """
+    Get deltas from a specific version of a dataset.
+    """
+    version = version.replace(settings.INDEX_VERSION, "")
```
SimonThordal, line 228: Should be factored out to avoid falling apart once we make a change to versions in some other place.

yente/data/dataset.py
```diff
@@ -41,6 +45,8 @@ def __init__(self, catalog: DataCatalog["Dataset"], data: Dict[str, Any]):
                     ts = datetime_iso(mdt)
             self.version = iso_to_version(ts) or "static"
 
+        self.delta_index = data.get("delta_url", None)
```
pudo, line 48: Maybe self.delta_url? That might help keep it more explainable... the metadata is a mess anyways.

yente/data/dataset.py
```diff
@@ -12,13 +13,16 @@
...
+from yente.data.loader import load_json_url
+from yente import settings
 
 log = get_logger(__name__)
 BOOT_TIME = datetime_iso(datetime.utcnow())
 
 
 class Dataset(NKDataset):
     def __init__(self, catalog: DataCatalog["Dataset"], data: Dict[str, Any]):
+        self._available_versions_map: Dict[str, str] = {}
```
pudo, line 25: I am almost wondering if this wants to be a separate object - a delta ops iterator (maybe a DatasetDelta?) that knows the dataset and the version in the index and then does all the handling?

yente/data/loader.py
```diff
@@ -36,13 +36,14 @@ async def load_json_url(url: str) -> Any:
     else:
         async with httpx_session() as client:
             resp = await client.get(url)
+            resp.raise_for_status()
```
pudo, line 39: oops :( that should have been there

yente/search/base.py
```diff
@@ -74,3 +91,317 @@ async def close_es() -> None:
...
+        )
+
+    async def _add_write_block(self, index: str) -> None:
+        await self.client.indices.put_settings(
+            index=index, settings={"index.blocks.read_only": True}
+        )
+
+    async def clone_index(self, index: str, new_index: str) -> None:
+        try:
+            await self._add_write_block(index)
```
pudo, line 221: I wonder if the write block should be added after the index is written, not before it is cloned.

yente/search/base.py
```diff
@@ -74,3 +91,317 @@ async def close_es() -> None:
...
+        return False
+
+    async def rollover(self, alias: str, new_index: str, prefix: str = "") -> None:
+        """
+        Remove all existing indices with a given prefix from the alias and add the new one.
+        """
+        actions = []
+        actions.append({"remove": {"index": f"{prefix}*", "alias": alias}})
+        actions.append({"add": {"index": new_index, "alias": alias}})
+        await self.client.indices.update_aliases(actions=actions)
```
pudo, line 259: Making this fully atomic is hella cool

yente/search/base.py
```diff
@@ -74,3 +91,317 @@ async def close_es() -> None:
...
+            "_op_type": "delete",
+            "_index": index,
+            "_id": doc_id,
+        }
+
+    def _update_operation(
+        self, entity: Dict[str, Any], doc_id: str, index: str
+    ) -> Dict[str, Any]:
+        return {
+            "_op_type": "update",
```
pudo, line 309: Are we sure we want to update and not index here?

yente/search/indexer.py
```diff
@@ -76,7 +85,7 @@ async def index_entities_rate_limit(
 async def index_entities(es: AsyncElasticsearch, dataset: Dataset, force: bool) -> bool:
     """Index entities in a particular dataset, with versioning of the index."""
     if not dataset.load:
-        log.debug("Dataset is going to be loaded", dataset=dataset.name)
+        log.debug("Dataset is not going to be loaded", dataset=dataset.name)
```
pudo, line 88: lol --- thanks for fixing it all up :)

yente/search/indexer.py
pudo: +1

yente/search/indexer.py
pudo: Next PR :)

yente/search/indexer.py
```diff
@@ -202,3 +214,105 @@ async def update_in_thread() -> None:
...
+    clone = None
+    try:
+        current_version = await get_current_version(dataset, provider)
+        if current_version is None:
+            raise Exception("No index found for dataset.")
+        index = Index(provider, dataset.name, current_version)
+        target_version = await dataset.newest_version()
+        # If delta versioning is not implemented, update the index from scratch.
+        if target_version is None:
+            raise DeltasNotAvailable(f"No versions available for {dataset.name}")
```
pudo, line 279: Is this where it'd branch out into a full re-index?

pudo (commented):
Really enjoy what's happening here :)

tests/unit/test_dataset.py
SimonThordal, line 1: Delete me

yente/data/dataset.py
```diff
@@ -72,3 +113,7 @@ def to_dict(self) -> Dict[str, Any]:
         if "children" not in data:
             data["children"] = [c.name for c in self.children]
         return data
+
+
+def get_delta_version(dataset_name: str, version: str) -> str:
+    return f"https://data.opensanctions.org/artifacts/{dataset_name}/{version}/entities.delta.json"
```
SimonThordal, line 119: Not used

yente/search/base.py
SimonThordal: ES won't let you clone the index if it isn't read only, but it's a good idea to make 100% sure there is no missed writes on the clone by doing the rollover before we remove the write block. So it's bee 
1. Add write block
2. Clone index
3. Update clone
4. Rollover to clone
5. Delete the old index
If there is an exception, delete the clone and remove the write block from the original index.

yente/search/base.py
SimonThordal: Yeah, it's a cool way of doing index migration tbh

yente/search/base.py
SimonThordal: We can use index as well as long as we're sure we'll never ship partial entities. Then we can combine _create and _update into _upsert.

yente/search/indexer.py
SimonThordal: Haha, I spent a good 10 minutes being very confused by our log messages :D

yente/search/indexer.py
SimonThordal: Yep, any exception should cause a full re-index.

yente/data/dataset.py
SimonThordal: It started out like that actually, but I moved away from it as complexity grew. I'll give it another spin and see if it makes sense though :)

docker-compose.testing.yml
SimonThordal, line 1: This was created because `action.destructive_requires_name=false` lets you wildcard in deletes. Super useful for cleaning up tests, but not a good production setting.

SimonThordal, line 1: I can just change this in the workflow file

PR #478, opened 2024-07-03
feat: add dummy authentication

SimonThordal:


pudo:
Nice work! I think one test we should do is if basic auth isn't already supported by embedding the credentials in the URL. In that case it may make more sense as a contingency plan to make a custom value for the Auth header, along the lines of what our own API does (`Authorization: Token XXXX`)?

PR #479, opened 2024-07-09
Experiment with separating provider and delta problems

pudo:
I'm wondering if we do need to separate these two issues after all. This is a little sketch implementation - untested and super hacky - that just does the deltas procedurally. WIP and discussion kick-starter.


yente/data/delta.py
```diff
@@ -0,0 +1,108 @@
...
+                dataset=self.dataset.name,
+            )
+            return False
+        if self.force_full:
+            return True
+        if self.target_version == self.base_version:
+            return False
+        return True
+
+    async def load(self) -> AsyncGenerator[EntityOp, None]:
```
SimonThordal, line 95: Nice, this is a cool unified way of dealing with this.

yente/search/indexer.py
```diff
@@ -75,6 +87,86 @@ async def iter_entity_docs(
...
+    except NotFoundError:
+        return None
+    versions: List[str] = []
+    for index in resp.keys():
+        ds, version = parse_index_name(index)
+        if ds == dataset.name:
+            versions.append(version)
+    if len(versions) == 0:
+        return None
+    # Return the oldest version of the index. If multiple versions are linked to the
```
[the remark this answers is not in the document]
pudo: Hmmmm I think (!) it's one index per dataset, but multiple datasets?

yente/data/delta.py
```diff
@@ -0,0 +1,108 @@
...
+log = get_logger(__name__)
+
+
+class DeltaIndex(TypedDict):
+    versions: Dict[str, str]
+
+
+class EntityOp(TypedDict):
+    op: str
+    entity: Dict[str, Any]
```
SimonThordal, line 17: :100:

yente/data/delta.py
```diff
@@ -0,0 +1,108 @@
...
+        return loader
+
+    @property
+    def is_incremental(self) -> bool:
+        """Check if there is sequence of delta entity patches that can be loaded."""
+        if self.force_full:
+            return False
+        return self.delta_urls is not None and len(self.delta_urls) > 0
+
+    def check(self) -> bool:
```
SimonThordal, line 79: Rename to `needs_update`? `check` makes it a bit hard to parse from the callsite.

yente/search/indexer.py
```diff
@@ -32,27 +33,38 @@
...
     datasets = set(dataset.dataset_names)
     idx = 0
-    async for data in load_json_lines(dataset.entities_url, index):
+    ops: Dict[str, int] = {"ADD": 0, "DEL": 0, "MOD": 0}
+    async for data in loader.load():
         if idx % 1000 == 0 and idx > 0:
             log.info("Index: %d entities..." % idx, index=index)
+        op_code = data["op"]
         idx += 1
+        ops[op_code] += 1
```
SimonThordal, line 57: Here I think it's better to keep the construction of the EntityOp in the SearchProvider or this will become a mess the day OpenSearch and ElasticSearch diverge.

yente/data/delta.py
pudo: Very very true!

yente/search/indexer.py
```diff
@@ -75,6 +87,89 @@ async def iter_entity_docs(
...
+    for index in resp.keys():
+        try:
+            ds, version = parse_index_name(index)
+            if ds == dataset.name:
+                versions.append(version)
+        except ValueError:
+            pass
+    if len(versions) == 0:
+        return None
+    # Return the oldest version of the index. If multiple versions are linked to the
```
SimonThordal, line 169: This is essentially everything in the Index class, right? If it takes the SearchProvider class instead then we'll save ourselves a headache when adding OpenSearch support.

yente/search/indexer.py
pudo: I feel like the EntityOp is sort of the interface between the loader and the provider - then turning it into ES/OS ops is where it becomes engine-specific?

yente/data/delta.py
```diff
@@ -0,0 +1,108 @@
...
+        self, dataset: Dataset, base_version: Optional[str], force_full: bool = False
+    ) -> None:
+        self.dataset = dataset
+        self.target_version = dataset.version or "static"
+        self.base_version = base_version
+        self.force_full = force_full
+        self.delta_urls: Optional[List[Tuple[str, str]]] = None
+
+    @classmethod
+    async def build(
```
SimonThordal, line 34: What do you say to the loader taking a provider so that it can produce operations based on what we have in the delta files?

pudo: Do you mean have the loader do what's in `yente.search.base:make_indexable` and the other iter_entity_docs function? Feels a bit like an overstepping of the data layer into backend concerns to me.

yente/search/indexer.py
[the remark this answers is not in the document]
SimonThordal: Sorry, I missed that you are doing a check on the dataset name further up. All good, nothing to see here :)

yente/search/indexer.py
```diff
@@ -123,23 +197,14 @@ async def index_entities(es: AsyncElasticsearch, dataset: Dataset, force: bool)
...
-        )
+    if updater.is_incremental and not force:
+        base_index = construct_index_name(dataset.name, updater.base_version)
+        await clone_index(es, base_index, next_index)
+    else:
+        await create_index(es, next_index)
 
     try:
-        docs = iter_entity_docs(dataset, next_index)
+        docs = iter_entity_docs(updater, next_index)
```
SimonThordal, line 207: This is great tbh, so much easier than the juggling in the other PR

yente/data/delta.py
SimonThordal: You might be right there. We could also give a SearchProvider to `iter_entity_docs` so that it is able to take the JSON from the loader and turn it into something that can be passed to ES/OS as an operation.

PR #489, opened 2024-07-30
feat: add traceID and pass it to ES

SimonThordal:
Adds trace IDs to requests as based on https://www.w3.org/TR/trace-context/#processing-model and passes them to ES. The major win here is that we're not breaking traces if we're part of a clients distributed stack.

We're not fully compliant as we're not doing full validation of the incoming traces or inspecting the flags, but we currently correctly participate in or initiate traces and add the trace context to our logs. ES also emits the trace context in logs, but it does not seem like OS [supports it quite yet](https://opensearch.org/docs/latest/observing-your-data/trace/distributed-tracing/).

I removed our trace ID implementation completely, but we could also deprecate it.

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -7,14 +7,12 @@ def test_healthz():
     res = client.get("/healthz")
     assert res.status_code == 200, res
     assert res.json().get("status") == "ok", res
-    assert "x-trace-id" in res.headers
 
 
 def test_readyz():
     res = client.get("/readyz")
     assert res.status_code == 200, res
     assert res.json().get("status") == "ok", res
-    assert "x-trace-id" in res.headers
 
 
 def test_manifest():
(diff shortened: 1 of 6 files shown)


pudo:
Wow, that's quite a chunk of code. Would it make sense to put in a few unit tests for the tracer classes? They're the type of thing that is going to remain untouched for years after we merge this until someone points out that it hasn't worked since three releases....

SimonThordal:
> Wow, that's quite a chunk of code. Would it make sense to put in a few unit tests for the tracer classes? They're the type of thing that is going to remain untouched for years after we merge this until someone points out that it hasn't worked since three releases....

Sure, I have a Postman test suite that I'm using for it, but we can add them as integration tests as well.

SimonThordal:
> Wow, that's quite a chunk of code. Would it make sense to put in a few unit tests for the tracer classes? They're the type of thing that is going to remain untouched for years after we merge this until someone points out that it hasn't worked since three releases....

Now with a test.

pudo (approved):
Nice work!

PR #502, opened 2024-08-14
Retry full on errors

SimonThordal:
Here's a suggestion for how to deal with the error reporting.
We should:
- Always retry on errors ( this was supposed to already have been in place, so not getting it out the first time around was a mistake )
- When individual documents in a bulk op fails we should log the error, not raise an exception. The update is still seen as successful. 

Not sure if we want to be stricter on the individual docs failing?

### DIFF
--- a/yente/provider/elastic.py
+++ b/yente/provider/elastic.py
@@ -229,28 +229,23 @@ class ElasticSearchProvider(SearchProvider):
             )
             raise YenteIndexError(f"Could not search index: {ae}") from ae
 
-    async def bulk_index(
-        self, entities: AsyncIterator[Dict[str, Any]]
-    ) -> Tuple[int, int]:
+    async def bulk_index(self, entities: AsyncIterator[Dict[str, Any]]) -> int:
         """Index a list of entities into the search index.
 
         Args:
             entities: An async iterator of entities to index.
 
         Returns:
-            A tuple of the number of entities indexed and the number of errors.
+            he number of entities indexed
         """
         try:
-            n, errors = await async_bulk(
+            n, _ = await async_bulk(
                 self.client(),
                 entities,
                 chunk_size=1000,
                 raise_on_error=False,
             )
-            errors = cast(List[Any], errors)
-            for error in errors:
-                log.error("Bulk index error", error=error)
-            return n, len(errors)
+            return n
 
         except BulkIndexError as exc:
             raise YenteIndexError(f"Could not index entities: {exc}") from exc
--- a/yente/provider/elastic.py
+++ b/yente/provider/elastic.py
@@ -1,7 +1,7 @@
 import json
 import asyncio
 import warnings
-from typing import Any, Dict, List, Optional, cast
+from typing import Any, Dict, List, Optional, cast, Tuple
 from typing import AsyncIterator
 from elasticsearch import AsyncElasticsearch, ElasticsearchWarning
 from elasticsearch.helpers import async_bulk, BulkIndexError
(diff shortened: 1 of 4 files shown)


yente/provider/elastic.py
```diff
@@ -229,15 +229,28 @@ async def search(
...
+        """
         try:
-            await async_bulk(
+            n, errors = await async_bulk(
                 self.client(),
                 entities,
                 chunk_size=1000,
-                yield_ok=False,
-                stats_only=True,
+                raise_on_error=False,
```
pudo, line 248: Does this mean that when there is an error, the indexer doesn't abort and still streams in a full set of entities? I kind of want to think through whether that has a way of backfiring (not obvious, but it is a change).

yente/search/indexer.py
```diff
@@ -162,18 +159,24 @@ async def index_entities(
...
     await provider.rollover_index(
         alias,
         next_index,
         prefix=dataset_prefix,
     )
+    # Delete any old indices that are no longer aliased to the current index.
+    for index in await provider.get_all_indices():
+        if index.startswith(dataset_prefix) and index != next_index:
+            log.info("Deleting old index", index=index)
+            await provider.delete_index(index)
```
pudo, line 177: I think we have that covered in the cleanup function in the same module - it's just moved location.

pudo (commented):
Looks simple enough, two quick checkins :)

pudo:
Ah one more question: do we have a way of returning a non-zero exit code from the `yente reindex` thing if there was an error? That way we can trigger alerts in the infrastructure more easily...

yente/provider/elastic.py
SimonThordal: That's what it means, yes - I made the change because we have stats_only set to True which doesn't do anything if raise_on_error isn't False. 

The logs shared in Slack show that some users are having issues with single records failing, so currently that would mean that the entire update (full or delta, doesn't matter) will fail and they will be stuck on whatever version they were on before. 
This way an error gets logged and the update succeeds partially which seems to happen quite a bit.
I see two potential issues with this:
1. The onus is on us to discover the single document errors and fix them. Since we don't have access to client logs we'd need to discover it in the logs of the ServiceAPI. This will be the case no matter what approach we take, but failing with no updates is a lot louder than partially succeeding.
2. There is an error in a delta update that would have worked if run with a full re-index. In this case they will be missing out on that operation and the error will be propagated forward since they're on delta updates.

As an aside, maybe we should have a "turn it off and on again" schedule? Every X days we do a full reindex instead of a delta update.

yente/search/indexer.py
SimonThordal: Cool, I'll just delete the todo then.

yente/search/indexer.py
```diff
@@ -205,11 +197,18 @@ async def update_index(force: bool = False) -> bool:
         catalog = await get_catalog()
         log.info("Index update check")
         changed = False
+        exceptions = []
         for dataset in catalog.datasets:
-            _changed = await index_entities_rate_limit(provider, dataset, force)
-            changed = changed or _changed
-
+            try:
+                _changed = await index_entities_rate_limit(provider, dataset, force)
```
SimonThordal, line 203: @pudo  Untested, but if we want to return an error from the process and still update each index possible then maybe something like this? If there's an error in an index the information about whether there was a change will be lost, but that might be OK if we don't need that level of granulatiry.

yente/provider/elastic.py
pudo: If a single record fails to index, the index job has failed and we shouldn't migrate to that version. Anything else makes this software a liability nightmare for us and our customers. Can we please make sure that principle is restored and that there's no chance ever that a non-complete index is rolled into? 

Regarding the "lets do a full reindex every Nth time" - I like that, I just don't know where we'd store it. Doing a full special index for it would be a bit much.

yente/search/indexer.py
pudo: I wonder if we should have this return a named tuple that contains: `failed`/`changed` as two different return values -

SimonThordal:
In the end any errors on a dataset will now log an error and then re-index from full. 
Does not return the number of updated entities as we no longer need that for the serviceapi

PR #555, opened 2024-11-07
Allow passing custom HOST through an env var.

lbrito1:
Currently the sever is hardcoded to listen at IPv4 only (0.0.0.0); this is a problem when running on certain infrastructure such as Fly.io, where the default private network is IPv6-only. By allowing users to pass a custom HOST, they can pass "::" to use IPv6.


pudo:
Very nice path, appreciate it!

PR #618, opened 2024-12-18
fix(mapping): Allow to auto expanding replica count

faishal:
## Problem

The fixed replica mapping with `auto_expand_replicas` is causing issues in our AWS OpenSearch instance configured with Zone Awareness. During cron job executions, new index creation fails because shard allocation does not comply with the [Zone Awareness settings](https://aws.amazon.com/blogs/big-data/impact-of-infrastructure-failures-on-shard-in-amazon-opensearch-service/).

## Solution

AWS recommends configuring a proper replica count to ensure shard allocation works seamlessly with Zone Awareness. This PR introduces the ability to configure the `auto_expand_replicas` setting in the index mapping, allowing more flexibility and resolving shard allocation failures.


pudo:
Thank you very much for taking the time to make a PR for this. Looks good, I'll merge an make a patch release.

PR #658, opened 2025-02-04
Dockerfile: Use multi-stage build and base it on slim python image 

legal90:
This Pull Request updates Dockerfile with the following changes:
1) Split it to use a multi-stage build, so the result image contain the bare minimum for running the application (Python interpreter and virtualenv with packages installed)
2) Switch to the smaller image `python:3.12-slim` (based on `debian:bookworm-slim`) to reduce the image size and reduce the attack surface.

### Image size:
  - Before: 1.23 GiB
  - After: 712 MiB

### Vulnerabilities:
  - Before: 803
  - After: 41
_(* Analyzed by GCP Container Analysis API)_

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,35 +1,43 @@
-FROM ubuntu:24.04
-ENV DEBIAN_FRONTEND=noninteractive
+# Define the base image to use for both stages
+ARG python_image=python:3.12-slim
+
+# ------------------------------------------------------------------------
+# Stage 1: Build and install dependencies
+# ------------------------------------------------------------------------
+FROM ${python_image} AS build
+
+# Install build dependencies
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    build-essential \
+    pkg-config \
+    libicu-dev \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install python package dependencies
+ENV PATH="/opt/venv/bin:$PATH"
+COPY . /app
+RUN python -m venv /opt/venv
+RUN pip install --no-cache-dir -e /app
+
+# ------------------------------------------------------------------------
+# Stage 2: Export a final runtime image
+# ------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


jbothma:
This looks really nice, thank you! Taking a closer look

Dockerfile
```diff
@@ -1,35 +1,46 @@
...
-    && ln -fs /usr/share/zoneinfo/Etc/UTC /etc/localtime \
-    && dpkg-reconfigure -f noninteractive tzdata \
-    && groupadd -g 10000 -r app \
-    && useradd -m -u 10000 -s /bin/false -g app app
+# Set environment variables
+ENV LANG="en_US.UTF-8"
+ENV TZ="UTC"
+ENV PYTHONUNBUFFERED=1
+ENV VIRTUAL_ENV=/opt/venv
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
```
jbothma, line 38: Libicu is also needed in the final image

```suggestion
ENV PATH="$VIRTUAL_ENV/bin:$PATH"

+RUN apt-get update && apt-get install -y libicu72 \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*

```

Dockerfile
```diff
@@ -1,35 +1,46 @@
...
+# ------------------------------------------------------------------------
+FROM ${python_image}
 
 LABEL org.opencontainers.image.title="OpenSanctions yente"
 LABEL org.opencontainers.image.licenses="MIT"
 LABEL org.opencontainers.image.source="https://github.com/opensanctions/yente"
 
-RUN apt-get -qq -y update \
-    && apt-get -y upgrade \
-    && apt-get -y install locales ca-certificates tzdata curl python3-pip \
```
jbothma, line 10: I think we also don't want to lose `ca-certificates`

jbothma:
Thanks, merged in #670

legal90:
Thank you, @jbothma !

PR #704, opened 2025-03-31
Restore old behavor of match API where topics are OR'd

leonhandreke:
This got changed by https://github.com/opensanctions/yente/pull/648.


pudo:
Would it be more consistent to expose this as a query parameter like we did for the /search API? Just want to avoid user vertigo here.

leonhandreke:
Hm, honestly, I think the filter_op on search is really confusing. Like, what is the difference between `exclude_dataset`, `include_dataset`, `dataset` and which of these does `filter_op` apply to (we should update the docs on that, I wouldn't be able to tell you and we touched the code just a couple weeks ago).

`/match` only has `topics` as a filter param, and making those AND makes no sense. So I'd vote to keep it like this.

I'll merge this in to get it out today, let's talk about if offline if you think this warrants more thought :)

PR #705, opened 2025-03-31
Add add docs for adjacent entities endpoints and make them visible

jbothma:
image

### DIFF
--- a/yente/routers/search.py
+++ b/yente/routers/search.py
@@ -229,7 +229,7 @@ async def fetch_entity(
 async def fetch_adjacent_entities(
     response: Response,
     entity_id: str = Path(
-        description="ID of the entity whose peers are requested", examples=["Q7747"]
+        description="ID of the entity whose graph context was requested", examples=["Q7747"]
     ),
     provider: SearchProvider = Depends(get_provider),
     sort: List[str] = Query([], title="Sorting criteria"),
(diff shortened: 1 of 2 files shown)


yente/routers/search.py
```diff
@@ -217,19 +217,19 @@ async def fetch_entity(
...
         404: {"model": ErrorResponse, "description": "Entity not found"},
         500: {"model": ErrorResponse, "description": "Server error"},
     },
-    include_in_schema=False,  # Hide while we try it out.
 )
 async def fetch_adjacent_entities(
     response: Response,
     entity_id: str = Path(
-        description="ID of the entity to retrieve", examples=["Q7747"]
+        description="ID of the entity whose peers are requested", examples=["Q7747"]
```
pudo, line 232: whose graph context was requested? peers introduces a new technical term we haven't used yet.

yente/routers/search.py
```diff
@@ -242,6 +242,15 @@ async def fetch_adjacent_entities(
         0, title="Start at result with given offset", le=settings.MAX_OFFSET
     ),
 ) -> Union[RedirectResponse, EntityAdjacentResponse]:
+    """***Beta:** This endpoint is released for wider testing and is not yet recommended
```
pudo, line 245: Remove one *

jbothma: this is for the * reference by the endpoint names in the menu, for people who might see the name and signature but not read the bold beta in the description. Is that overkill?

PR #718, opened 2025-04-04
Fix adjacent expansion when root is interstitial and property name conflicts in mapping

jbothma:
The adjacent API currently doesn't expand outgoing edges when the queried entity is an interstitial entity. This is because the logic for initial outgoing props was using the prop schema type rather than the prop type by mistake. The intention there is to only query for values that are entity IDs - I think I copied logic for expanding to edges when that's totally irrelevant there.

buggy

image

less buggy

image

---

image

### DIFF
--- a/tests/test_entity.py
+++ b/tests/test_entity.py
@@ -66,6 +66,19 @@ def test_entity_nested():
     assert cdu["schema"] == "Organization"
     assert cdu["properties"]["name"] == ["CDU"]
 
+    # Nested on an interstitial entity
+    res = client.get(f"/entities/{pmt['id']}")
+    assert res.status_code == 200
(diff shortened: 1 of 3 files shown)


leonhandreke:
I think I understand what's happening and this looks good to me.

I don't think it would hurt to add a comment explaining this succinct difference, also below in the iteration where the check for `adj.schema.edge` is essential to terminating the loop

pudo (approved):
Thank you for improving the test coverage around this. One day we really should build a perfect FtM test dataset with all the graphy and property-y things. 

The code improvements look great IMO and I'm more and more happy with plan C on the prop overlaps. Seems non-intrusive.

PR #779, opened 2025-06-20
search: dial down length normalization on names & up-score large docs

leonhandreke:
Brought to you by the Friday-afternoon labs of search engine
engineering! The end result here is that Vlad is first for the query
"Putin", and that Daniel Putin, the former Mayor of Champagnat, is a bit
further down.

Length normalization on names is a bit of a difficult topic for us:
Important entities for us usually have a lot of names, and if only some
of them match, we shouldn't downrank it too much for having lots of others.
This is the reason why currently, small docs with one exact name match
always come up first. However, we also have to be careful not to rank
"Putin’s chef" too highly for the "Putin" query, so we don't dial it all
the way down to 0.

This commit also introduces a num_values field in the index that we use
to up-score "important" docs a bit.

### DIFF
--- a/yente/search/search.py
+++ b/yente/search/search.py
@@ -75,16 +75,10 @@ def result_facets(
     return facets
 
 
-async def search_entities(
-    provider: SearchProvider,
-    query: Dict[str, Any],
-    limit: int = 5,
-    offset: int = 0,
-    aggregations: Optional[Dict[str, Any]] = None,
-    sort: List[Any] = [],
-) -> Dict[str, Any]:
-    # Wrap query in function_score to up-score important entities.
-    function_score_query = {
+def upscore_large_entities(query: Dict[str, Any]) -> Dict[str, Any]:
+    """Wrap query to up-score important entities."""
+
+    return {
         "function_score": {
             "query": query,
             "functions": [
@@ -106,9 +100,18 @@ async def search_entities(
         }
     }
 
+
+async def search_entities(
+    provider: SearchProvider,
+    query: Dict[str, Any],
+    limit: int = 5,
+    offset: int = 0,
+    aggregations: Optional[Dict[str, Any]] = None,
+    sort: List[Any] = [],
+) -> Dict[str, Any]:
     return await provider.search(
         index=settings.ENTITY_INDEX,
-        query=function_score_query,
+        query=query,
         size=limit,
         sort=sort,
         from_=offset,
--- a/yente/routers/search.py
+++ b/yente/routers/search.py
@@ -17,7 +17,7 @@ from yente.provider import SearchProvider, get_provider
 from yente.search.queries import parse_sorts, text_query
 from yente.search.queries import facet_aggregations
 from yente.search.queries import Filters, Operator
-from yente.search.search import get_entity, search_entities
+from yente.search.search import get_entity, search_entities, upscore_large_entities
 from yente.search.search import result_entities, result_facets, result_total
 from yente.search.nested import get_nested_entity, get_adjacent_entities
 from yente.data import get_catalog
@@ -152,6 +152,7 @@ async def search(
         filter_op=filter_op,
     )
     aggregations = facet_aggregations([f.value for f in facets])
+    query = upscore_large_entities(query)
     resp = await search_entities(
         provider,
         query,


leonhandreke:
Note that the length norm change in `mapping.py` will also affect scoring for match candidate retrieval. I think the change in `search.py` only affects search.

yente/search/indexer.py
```diff
@@ -64,6 +64,9 @@ async def iter_entity_docs(
             doc = entity.to_full_dict(matchable=True)
             entity_id = doc.pop("id")
             doc["entity_id"] = entity_id
+            # Total number of values in the entity, used to up-score on
+            # large (i.e. important) entities.
+            doc["num_values"] = sum([len(v) for v in doc["properties"].values()])
```
jbothma, line 69: there are also potential differences in entity types. E.g. we generally have very few properties for securities

yente/search/search.py
```diff
@@ -85,9 +85,29 @@ async def search_entities(
...
+                        "field": "num_values",
+                        # This is a bit of a jiggle factor. Currently, very large documents (like Vladimir Putin)
+                        # have a num_values of ~200, so get a +10 boost.
+                        # The order is modifier(factor * value)
+                        "factor": 0.5,
+                        "modifier": "sqrt",
+                    }
+                }
+            ],
+            "boost_mode": "sum",
```
jbothma, line 104: so 

1 prop value ends up doing subquery_score + 0.71
2 prop values end up doing subquery_score + 1
3 prop values end up doing subquery_score + 1.2

what sort of range would subquery_score be for a search for something like `John Smith`? How big an impact does it make for cases where lots of candidates with very similar scores?

I'm exploring what the chance is of us excluding candidates with very few props but that match decently, for things that elastic scores similarly but for whom a boost of 1 would be a big deal.

jbothma (commented):
These look like very nice sensible changes.

My biggest worry here is that it's hard to know whether/how much we'll negatively affect candidate generation for matching.

Any thoughts how we can get a sense of that? I don't think we can store queries and compare results between main and this. We can generate a set of queries from data we have and compare the different results (limited to the ones that differ after scoring, perhaps).

yente/search/search.py
leonhandreke: So *the* Obama will get a +7.4, *the* Putin a +10.

yente/search/indexer.py
jbothma: or `* 1 + (len(topics)/10)` or something - not insane

yente/search/search.py
jbothma: ok - I think the relative size between subquery_Score and what we're adding is sane

yente/search/mapping.py
```diff
@@ -30,6 +30,15 @@
         "refresh_interval": "5s",
         "auto_expand_replicas": settings.INDEX_AUTO_REPLICAS,
         "number_of_shards": settings.INDEX_SHARDS,
+        "similarity": {
+            # We use this for names, to avoid over-penalizing entities with many names.
+            "weak_length_norm": {
+                # BM25 is the default similarity algorithm.
+                "type": "BM25",
+                # 0.75 is the default
+                "b": 0.25,
```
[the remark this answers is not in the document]
jbothma: this is super useful - what do you think of adding pretty much your comment here as a comment in the code so we have an idea how we came to this number if we want to tweak it in future?

jbothma:
I think the sort of place where bad things might happen is with really common names. Perhaps another good dataset to test with is us_congress.

Then I think it's worth trying!

leonhandreke:
```
                Algorithm Performance Comparison
                        Dataset: default
                        Matcher: logic-v1
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Person File                         ┃ leonhandreke/score-size ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ positives_un_untreated.csv          │         1.000 (100/100) │
│ positives_us_congress_untreated.csv │         1.000 (100/100) │
│ negatives_us.csv                    │          0.438 (20/100) │
└─────────────────────────────────────┴─────────────────────────┘

                Algorithm Performance Comparison
                        Dataset: default
                   Matcher: UNSTABLE-logic-v2
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Person File                         ┃ leonhandreke/score-size ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ positives_un_untreated.csv          │         0.983 (100/100) │
│ positives_us_congress_untreated.csv │         0.978 (100/100) │
│ negatives_us.csv                    │          0.389 (10/100) │
└─────────────────────────────────────┴─────────────────────────┘
```

```
            Algorithm Performance Comparison
                    Dataset: default
                    Matcher: logic-v1
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┓
┃ Person File                         ┃         develop ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
│ positives_un_untreated.csv          │ 1.000 (100/100) │
│ positives_us_congress_untreated.csv │ 1.000 (100/100) │
│ negatives_us.csv                    │  0.423 (20/100) │
└─────────────────────────────────────┴─────────────────┘

            Algorithm Performance Comparison
                    Dataset: default
               Matcher: UNSTABLE-logic-v2
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┓
┃ Person File                         ┃         develop ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
│ positives_un_untreated.csv          │ 0.983 (100/100) │
│ positives_us_congress_untreated.csv │ 0.977 (100/100) │
│ negatives_us.csv                    │  0.406 (10/100) │
└─────────────────────────────────────┴─────────────────┘
```

Only difference is in the true-negatives, which have very common names. Not sure if one could interpret that this change agrees slightly more with logic-v2 or if this is statistical jitter/the numbers are too low.

leonhandreke:
Alright, this has been lingering for a while and we're wrapping up Yente 5. I'm reasonably confident that changing the length norm on names doesn't make candidate generation worse and is the correct thing to do given the type of data we're searching.

I've moved the up-scoring of large docs out of `search_entities` so that it only kicks in for `/search` - I think that's the right thing to do, since in `/match`ing, we don't care about having Putin first, we just want to find the needle in the haystack most accurately.

We've discussed a couple of alternatives to upscoring large docs (notably, upscoring on `len(topics)`), but I don't think anyone had strong opinions and I haven't really seen an argument why number of values on the entity isn't a good measure. And since it only kicks in for `/search` anyway, we can also change it if it turns out to be dumb (IIUC, we don't offer any guarantees on `/search`).

So unless there are major objections, I'll merge this to yente-next today.

Issue #769, opened 2025-06-10
Allow ignoring entry ids in the match endpoint


baldurh:
I would like to pass a list of entry ids to the match endpoint so that those entry ids are not fetched from ElasticSearch.

Our use case is as follows:

If we get a hit on an entity we store it in our system and it needs to be manually reviewed by a person. The hit can be whitelisted by them. Once a hit has been whitelisted we want to essentially ignore it and not consider it again.

We have some entities with very common names, and since there’s a limit on how many results are returned we usually get back the same handful of hits—which have already been whitelisted—when we screen them.

So we would like to be able to pass the list of known entry ids to the match endpoint so that we can be sure that our known hits do not "block" other hits.

baldurh:
> One detail that I'd appreciate your input on is how to deal with non-canonical IDs. Suppose you make a decision on an ID, and this entity later gets merged into another entity -- do you want that entity to remain excluded? My hunch would be to continue to exclude the whole cluster, because once you've decided that it's not that entity, you also don't care much if new information about that entity comes in, right? Would appreciate your input here! (I hope my question made sense, please let me know if you'd like more context).

@leonhandreke yes the entity should remain excluded even if it gets merged with another—ie. if the id is found in the list of referents it should not be included

leonhandreke:
Nice, thanks for the feedback! As mentioned, I'm hoping to get to this soon enough :)

PR #783, opened 2025-06-27
Add exclude_entity_ids param to /match

leonhandreke:
This will allow callers to exclude entities that have been marked as a
false-positive, possibly by a human. Passing this on to Elastic instead
of doing it client-side means we possibly get better match candidates.

See https://github.com/opensanctions/yente/issues/769

### DIFF
--- a/yente/routers/match.py
+++ b/yente/routers/match.py
@@ -66,6 +66,8 @@ async def match(
         pattern=TS_PATTERN,
         title="Match against entities that were updated since the given date",
     ),
+    # This list applies to all queries, but really only makes sense with a single query.
+    # Our API design currently doesn't support per-query options such as this one.
     exclude_entity_ids: List[str] = Query(
         [],
         title="A list of entities IDs to exclude from matching",


leonhandreke:
Logically, I think it really makes sense for this to go into a query and not into the request-global params (actually, a few of the params would really make sense on a per-query level). But unfortunately, the query is just an `EntityExample`, and we can't really stuff things in there...

yente/search/queries.py
```diff
@@ -82,6 +83,12 @@ def filter_query(
 
     for schema_name in exclude_schema:
         must_not.append({"term": {"schema": schema_name}})
+
+    # Exclude entities by any ID in the cluster
+    if exclude_entity_ids:
+        must_not.append({"terms": {"entity_id": exclude_entity_ids}})
```
jbothma, line 89: it feels like this maybe needs some kind of limit if passing in a massive list can cause elastic performance degradation. I guess if and what we limit this to depends on

- how many items a customer would normally want to supply here
- whether they can limit the numbers based on some kind of scope, e.g. false positives in Germany, or "false positives for Steve", or something
- if and at what point a massive list causes degradation on elastic

Just to throw out some kind of ballpark - I wouldn't be surprised if elastic wasn't too bothered by being supplied 100 keys in each of 5 requests per second. But would a customer be supplying around 50, or around 5000? And how does Elastic feel about 5000 in each request?

Either way, I think it's better to start with some limit even if it's too high for ES, but so that the idea of a limit is there from day 1 and not a big surprise for customers.

jbothma (approved):
I'm happy to be overridden but perhaps it's good to get a feel for how many ids customers anticipate supplying, and how many elasticsearch eats without blinking.

PR #788, opened 2025-07-07
Implement env var expansion for auth_token in manifest

leonhandreke:


yente/data/manifest.py
```diff
@@ -26,7 +28,10 @@ class CatalogManifest(BaseModel):
 
     async def fetch_datasets(self) -> List[Dict[str, Any]]:
         """Fetch the datasets from the catalog."""
-        data = await load_yaml_url(self.url, auth_token=self.auth_token)
+        auth_token_expanded = (
+            os.path.expandvars(self.auth_token) if self.auth_token else None
```
pudo, line 32: That feels like a very late moment to me to do this. Perhaps it could be a pydantic validator instead?

leonhandreke: You're right. Thank you.

PR #807, opened 2025-07-24
Google Cloud Logging compatible request logging

leonhandreke:
Google Cloud Logging has a magic httpRequest field in its LogEntry proto
that triggers special display logic (and more?). For all the people who
don't use Google Cloud, this isn't worse than what we had before. In
fact, it's better, case now we can process X-Fowarded-For.

### DIFF
--- a/yente/app.py
+++ b/yente/app.py
@@ -1,4 +1,3 @@
-import time
 import aiocron  # type: ignore
 from typing import AsyncGenerator, Dict, Type, Callable, Any, Coroutine, Union
 from contextlib import asynccontextmanager
@@ -10,7 +9,6 @@ from fastapi.middleware.cors import CORSMiddleware
 from fastapi.middleware.gzip import GZipMiddleware
 from starlette.middleware.base import RequestResponseEndpoint
 from fastapi.responses import JSONResponse
-from structlog.contextvars import clear_contextvars, bind_contextvars
 
 from yente import settings
 from yente.exc import YenteError
@@ -20,7 +18,7 @@ from yente.data import refresh_catalog
 from yente.data.entity import Entity
 from yente.search.indexer import update_index_threaded
 from yente.provider import close_provider
-from yente.middleware import TraceContextMiddleware
+from yente.middleware import RequestLogMiddleware, TraceContextMiddleware
 
 log = get_logger("yente")
 ExceptionHandler = Callable[[Request, Any], Coroutine[Any, Any, Response]]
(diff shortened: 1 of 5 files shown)


yente/middleware/request_log.py
```diff
@@ -0,0 +1,60 @@
...
+        }
+        if client_ip := get_client_ip(request):
+            http_request_info["remoteIp"] = client_ip
+
+        previous_contextvars = structlog.contextvars.bind_contextvars(
+            # This field is special in Google Cloud Logging, see
+            # https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest
+            httpRequest=http_request_info,
+            # Not part of the Google-spec'd httpRequest object, but we want to capture the request path for easier filtering.
+            requestPath=request.url.path,
```
jbothma, line 45: > Product-specific logging information MUST be defined in a separate message.

from https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest

Does this mean we shouldn't put custom stuff in his message? Can we link this message to the "separate message" with a tracing id? Not sure what the ux of that would be yet - still catching up to these PRs

jbothma (commented):
A change like this would affect teams with dashboards or other dependencies on the log format.

How about making this configurable as a list of loging middleware class paths, e.g.

`YENTE_LOG_MIDDLEWARE=yente.logging.BasicLogging,yente.logging.GoogleCloudLogging` to allow for a period of migration

leonhandreke:
Thinking about this some more, I feel like the format of the log output is something we should be able to change with a minor release and an item in the changelog? Sure, having a config variable and a 3-month migration window feels grown-up, but it also makes changes like this so much harder and less fun to execute :(

leonhandreke:
@jbothma How would you feel pulling this in to Yente 5.0? I still feel/want to believe that the content of log messages is not part of our guaranteed API (I don't think we document it anyway), but with 5.0, people should be paying attention anyway.

yente/app.py
```diff
@@ -44,32 +42,14 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
     await close_provider()
 
 
-async def request_middleware(
+async def catch_exception_middleware(
```
jbothma, line 45: just an idea, don't feel strongly about it

```suggestion
async def json_exception_middleware(
```

jbothma (commented):
Sorry looks like the latency comment didn't go out before.

Yeah I think for a major version bump it's cool. It'd at least be kind to put a note about the change in the release notes yeah.

PR #826, opened 2025-08-05
Add a pre-made delivery.opensanctions.com manifest

leonhandreke:
This allows people using the pre-built docker image to use delivery
service by setting the `YENTE_MANIFEST` and
`OPENSANCTIONS_DELIVERY_TOKEN` env vars.


pudo:
One thing we could do is to split the docker compose manifest: we have one configuration in `docker-compose.dev.yml` pointing to `manifests/dev.yml`, and one configuration in `docker-compose.commercial.yml` pointing to `manifests/commercial.yml` which already has a placeholder for the token?

docker-compose.commercial.yml
```diff
@@ -0,0 +1,16 @@
...
+      file: docker-compose.yml
+      service: index
+
+  app:
+    extends:
+      file: docker-compose.yml
+      service: app
+    environment:
+      YENTE_MANIFEST: "manifests/commercial.yml"
+      OPENSANCTIONS_DELIVERY_TOKEN: "your-secret-token"
```
jbothma, line 13: oh that's clever - using the environment variable so the file doesn't have to be mounted in.

when extending, is `environment` cumulative, or will this overwrite `environment in the original, meaning the YENTE_INDEX_URL setting gets lost?

Issue #821, opened 2025-08-04
Pull in nomenklatura 4.1


leonhandreke:
- [x] Make compatible with nomenklatura 4.0 https://github.com/opensanctions/yente/pull/798
- [x] Allow passing a `ScoringConfig` https://github.com/opensanctions/yente/issues/819
- [x] Config option to hide `logic-v2` from our production API https://github.com/opensanctions/yente/pull/832

leonhandreke:
This is done in `develop`, closing.

PR #832, opened 2025-08-07
Add YENTE_HIDDEN_ALGORITHMS env var to hide logic-v2 from prod for now

leonhandreke:
Question: Is the `assert` in `settings` a bit over the top? Should it be
an error/warning log in `lifespan` instead?

https://github.com/opensanctions/yente/issues/821

### DIFF
--- a/yente/routers/util.py
+++ b/yente/routers/util.py
@@ -36,12 +36,12 @@ ALGO_HELP = (
 
 def get_algorithm_by_name(name: str) -> Type[ScoringAlgorithm]:
     """Return the scoring algorithm class with the given name."""
-    name = name.strip()
-    if name == "best":
-        name = settings.BEST_ALGORITHM
+    name_clean = name.lower().strip()
+    if name_clean == "best":
+        name_clean = settings.BEST_ALGORITHM
     algorithm = None
     for a in ENABLED_ALGORITHMS:
-        if a.NAME == name:
+        if a.NAME == name_clean:
             return a
     if algorithm is None:
         raise HTTPException(400, detail=f"Invalid algorithm: {name}")
--- a/yente/routers/util.py
+++ b/yente/routers/util.py
@@ -33,6 +33,11 @@ ALGO_HELP = (
     f"Scoring algorithm to use, options: {ALGO_LIST} (best: {settings.BEST_ALGORITHM})"
 )
 
+# Ensure that all hidden algorithms are valid
+assert all(
+    algo in [a.NAME for a in ENABLED_ALGORITHMS] for algo in settings.HIDDEN_ALGORITHMS
+), "Invalid algorithm name in YENTE_HIDDEN_ALGORITHMS"
+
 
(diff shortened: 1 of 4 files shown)


yente/routers/util.py
```diff
@@ -1,31 +1,48 @@
...
+ALGO_LIST = ", ".join([a.NAME for a in ENABLED_ALGORITHMS])
 ALGO_HELP = (
     f"Scoring algorithm to use, options: {ALGO_LIST} (best: {settings.BEST_ALGORITHM})"
 )
 
 
 def get_algorithm_by_name(name: str) -> Type[ScoringAlgorithm]:
     """Return the scoring algorithm class with the given name."""
-    name = name.lower().strip()
+    name = name.strip()
```
jbothma, line 39: this might be a breaking change if someone's sending in `LOGIC-V1` and it's been working until now, right?

is this because the new algo is called `UNSTABLE-logic-v2`?

An alternative is to lowercase a.NAME for the comparison

yente/settings.py
```diff
@@ -153,6 +154,15 @@ def random_cron() -> str:
...
 DEFAULT_ALGORITHM = env_str("YENTE_DEFAULT_ALGORITHM", "logic-v1")
 BEST_ALGORITHM = env_str("YENTE_BEST_ALGORITHM", "logic-v1")
+HIDDEN_ALGORITHMS = [
+    algo for algo in env_str("YENTE_HIDDEN_ALGORITHMS", "").split(",") if algo != ""
+]
+# Ensure that all hidden algorithms are valid
+assert all(
+    algo in [a.NAME for a in nomenklatura.matching.ALGORITHMS]
+    for algo in HIDDEN_ALGORITHMS
+), "Invalid algorithm name in YENTE_HIDDEN_ALGORITHMS"
```
jbothma, line 164: I think this assertion is fair - better to find out that the algo we mean to hide isn't hidden - if we try to deploy an instance that doesn't start up, I think it won't scale down the old nodes. That's built into `dokku` as well, which gives me the idea that it can be expected from people's deployment processes.

yente/routers/util.py
leonhandreke: Yeah, this is breaking, I realized this too (but only put it in the body of the commit message).

I think it's sort of weird that it's case insensitive, and figured that now's a good time to break it.

But also there is not really any good reason to break it, other than that it's a bit weird.

WDYT?

jbothma: I think stricter interfaces are better than loosey goosey in a lot of cases. But I don't think breaking the API for someone without an announcement is nice. I assumed this is a change we're planning on releasing in saas soon so people can start trying the algo if they know about it. A surprise breaking change doesn't really fit with what we're trying to do here. So I think somehow supporting case insensitive algo selection as we had before is kinda needed if we don't announce and give people a chance to adapt.

jbothma (approved):
cool!

Issue #819, opened 2025-08-04
Support ScoringConfig in the /match API


pudo:
With `nomenklatura` 4.1, scoring systems can allow for fine tuning using a `ScoringConfig`. Part of that is the `weights`, which are already handled as part of the `/match` request body. Alongside `weights`, we should now also accept a `config` dict of [str, Any] to allow clients to rotate those knobs as needed. We should also test that the config works, and that config options are explained in the `/algorithms` endpoint.

PR #833, opened 2025-08-08
Implement passing scoring config in match request

leonhandreke:
Also, expose the options in the `/algorithms` endpoint.

Fixes https://github.com/opensanctions/yente/issues/819

### DIFF
--- a/yente/data/common.py
+++ b/yente/data/common.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 from typing import Any, Dict, List, Union, Optional
 from pydantic import BaseModel, Field
-from nomenklatura.matching.types import MatchingResult, FeatureDocs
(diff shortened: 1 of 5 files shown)


yente/data/common.py
```diff
@@ -121,6 +121,11 @@ class EntityExample(BaseModel):
 
 class EntityMatchQuery(BaseModel):
     weights: Dict[str, float] = Field({}, examples=[{"name_literal": 0.8}])
+    config: Dict[str, Union[str, int, float, bool]] = Field(
```
jbothma, line 124: Pydantic has this, in case we want to support nested config structures

```suggestion
    config: Dict[str, JsonValue] = Field(
```

jbothma: Oh I see this matches ConfigVar

jbothma (approved):
Nice! LGTM

Issue #806, opened 2025-07-24
Why are we indexing entity.to_full_dict?


leonhandreke:
In `iter_entity_docs`, we are calling `entity.to_full_dict`. This builds the inverted type groups, i.e. `names`, `countries`. This should already be covered by the `copy_to` magic in `make_type_field` in `mapping.py`. 

Figure out what's going on, fix it, move to `entity.to_dict`.

leonhandreke:
Little inline comment (forgot to link in commit message): https://github.com/opensanctions/yente/commit/e64bedc7b2551853f94bf686969947f7de748e12

leonhandreke:
Just some more context for code review: we're calling `Entity.to_full_dict` in the indexer, which builds all the FtM-type fields (like `countries` or `entities`) that should already be covered by the `copy_to` magic in Elastic (e.g. `copy_to: ["countries"]` on the `nationality` field). The reason for this is a bit unclear. It happened here: https://github.com/opensanctions/yente/commit/092dbb06d51ffb93afb713596b82975db49d9c18 and @pudo can't quite remember why (at least that's what I gathered from conversations). 

Moving to `Entity.to_dict` is good because a) it's a cleanup b) we might get some speed gains by sending only something like half the document size over the wire.

PR #839, opened 2025-08-13
Index Entity.to_dict, rely on copy_to magic, add tests for that

leonhandreke:
Previously, the NAME_{KEY,PART,PHONETIC} fields were built from the
`names` type field generated by `to_full_dict`. That's no longer being
called, so we're now pulling out the names using
`Entity.get_type_values` in `build_indexable_entity_doc` to do further
shenanigans on the names.

See https://github.com/opensanctions/yente/issues/806

### DIFF
--- a/tests/test_mappings.py
+++ b/tests/test_mappings.py
@@ -1,0 +1,91 @@
+import pytest
+
+from yente import settings
+from yente.search.mapping import INDEX_SETTINGS, make_entity_mapping
+from yente.data.entity import Entity
+from yente.search.indexer import build_indexable_entity_doc
+
+
+INDEX_MAPPINGS = make_entity_mapping()
+
+
+@pytest.mark.asyncio
+async def test_mappings_copy_to(search_provider):
+    """Test that all the mapping and indexing magic works.
+
+    We test that by executing queries on specific fields of the indexed documents."""
+    # Create a test index using the same settings as test_search_provider
+    temp_index = settings.ENTITY_INDEX + "-mappings-test"
+    try:
+        await search_provider.create_index(
+            temp_index, mappings=INDEX_MAPPINGS, settings=INDEX_SETTINGS
+        )
+
+        # Create a test entity with Vladimir Putin data
+        entity = Entity.from_dict(
+            {
+                "id": "Q7747",
+                "schema": "Person",
+                "properties": {
+                    "name": ["Vladimir Putin"],
+                    "citizenship": ["ru"],
+                    "topics": ["sanction"],
+                },
+                "datasets": ["test"],
+                "referents": [],
+                "first_seen": "2023-01-01T00:00:00",
+                "last_seen": "2023-01-01T00:00:00",
+                "last_change": "2023-01-01T00:00:00",
+            }
+        )
+
+        action = {
+            "_index": temp_index,
+            "_id": entity.id,
+            "_source": build_indexable_entity_doc(entity),
+        }
+        await search_provider.bulk_index([action])
+        await search_provider.refresh(temp_index)
+
+        # names is a text field, so need to use match
+        search_result = await search_provider.search(
+            temp_index, {"bool": {"must": [{"match": {"names": "Vladimir"}}]}}
+        )
+        assert len(search_result["hits"]["hits"]) == 1, "Failed to match on names"
+
+        # name_parts and name_phonetic are a bit of a special case, we syntesize them in the indexer
+        search_result = await search_provider.search(
+            temp_index, {"bool": {"must": [{"match": {"name_parts": "Vladimir"}}]}}
+        )
+        assert len(search_result["hits"]["hits"]) == 1, "Failed to match on name_parts"
+        search_result = await search_provider.search(
+            temp_index, {"bool": {"must": [{"match": {"name_phonetic": "FLTMR"}}]}}
+        )
+        assert (
+            len(search_result["hits"]["hits"]) == 1
+        ), "Failed to match on name_phonetic"
+
+        # Try to match on the countries field, which is a type field that is populated by copy_to from citizenship
+        search_result = await search_provider.search(
+            temp_index, {"bool": {"must": [{"term": {"countries": "ru"}}]}}
+        )
+        assert len(search_result["hits"]["hits"]) == 1, "Failed to match on countries"
+
+        # Try to match on the text field, which is a copy_to field that is populated by just about everything
+        search_result = await search_provider.search(
+            temp_index, {"bool": {"must": [{"term": {"text": "ru"}}]}}
+        )
+        assert (
+            len(search_result["hits"]["hits"]) == 1
+        ), "Failed to match country on text"
+        search_result = await search_provider.search(
+            temp_index, {"bool": {"must": [{"term": {"text": "sanction"}}]}}
+        )
+        assert len(search_result["hits"]["hits"]) == 1, "Failed to match topics on text"
+        search_result = await search_provider.search(
+            temp_index, {"bool": {"must": [{"match": {"text": "vladimir"}}]}}
+        )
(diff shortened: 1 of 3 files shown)


leonhandreke:
So I *think* that nothing is terribly broken, Elastic is working as intended (or rather, we are using it in the right way, I wouldn't have expected this to be broken). So I'm not sure why at some point we switched to `to_full_dict` instead of `to_dict`.

What I did discover is that the name processing shenanigans (phonetics and whatever `_KEY` is doing) rely on `names` being there in the dict, so I had to pull the names from the Entity using `Entity.get_type_values` instead.

I wrote a that ensures that three aspects of this work:

- `copy_to` for the type fields (like `countries`)
- `copy_to` for the `text` field
- The `name_parts` magic

The test just does search queries. Getting at the `doc_values` for keyword fields is a bit hard (would likely require changes to `SearchProvider`) and for `text` fields (like name) it seems to be impossible? Kibana does it somehow, but I think it uses some other magic APIs. Anyway, this is good enough to verify that it works I guess.

leonhandreke:
Following the merge copy_to commit, I did a diff of the old vs the new mapping. Only the order of the fields in the copy_to list has changed sometimes, everything else seems to be the same.

yente/search/mapping.py
```diff
@@ -91,12 +93,29 @@ def make_entity_mapping(schemata: Optional[Iterable[Schema]] = None) -> Dict[str
...
-                copy_to=copy_to,
-                existing=prop_mapping.get(name, None),
+            prop_name_to_fields[name].append(
+                make_type_field(prop.type, copy_to=copy_to)
             )
 
+    prop_mapping: Dict[str, MappingProperty] = {}
+    for prop_name, fields in prop_name_to_fields.items():
+        merged_copy_to = list(
+            set(itertools.chain.from_iterable([f["copy_to"] for f in fields]))
```
jbothma, line 103: omg they could have just called this flatten...
ok ok it's just one level and doesn't return a list but

yente/search/mapping.py
```diff
@@ -91,12 +93,29 @@ def make_entity_mapping(schemata: Optional[Iterable[Schema]] = None) -> Dict[str
...
+        keyword_fields = [f for f in fields if f["type"] == "keyword"]
+
+        # All fields within a group (text/keyword) are usually the same. We choose the
+        # last one cause that's what we used to do, but it really doesn't matter.
+        # keyword has precedence over text
+        # Currently, only authority causes a collision (some are string, some are entity)
+        selected_field = keyword_fields[-1] if keyword_fields else text_fields[-1]
+        # We merge the copy_to fields, just so that search still behaves as expected.
+        # even in the case of multiple fields with the same name.
+        selected_field["copy_to"] = merged_copy_to
```
jbothma, line 115: Just trying a different wording with a bit more precise terminology to see if I'm understanding correctly. No pressure to use this, but if I misunderstood, let me know please

```suggestion
        # All properties with the same name ought to map to the same field definition.
        # If a conflict occurs, we choose field type "keyword" over "text".
        # Currently, only the property "authority" has conflicts (some are string, some are entity)
        selected_field = keyword_fields[-1] if keyword_fields else text_fields[-1]
        # We merge the copy_to fields rather than choosing one value, just so that
        # queries work as expected in case one property maps to different copy_to than another
        # with the same field name.
        selected_field["copy_to"] = merged_copy_to
```

yente/search/mapping.py
```diff
@@ -76,7 +75,10 @@ def make_keyword() -> MappingProperty:
 def make_entity_mapping(schemata: Optional[Iterable[Schema]] = None) -> Dict[str, Any]:
     if schemata is None:
         schemata = list(model.schemata.values())
-    prop_mapping: Dict[str, MappingProperty] = {}
+    # Multiple schemata can have the same property name, but we flatten them
+    # into a single field in the search index. That's why we collect a list of
+    # fields for each property name first and resolve them later.
+    prop_name_to_fields: Dict[str, List[MappingProperty]] = defaultdict(list)
```
jbothma, line 81:
```suggestion
    # Collect field definitions:
    # Multiple schemata can have the same property name, but we flatten them
    # into a single field in the search index. That's why we collect a list of
    # fields for each property name first and resolve them later.
    prop_name_to_fields: Dict[str, List[MappingProperty]] = defaultdict(list)
```

yente/search/mapping.py
```diff
@@ -91,12 +93,29 @@ def make_entity_mapping(schemata: Optional[Iterable[Schema]] = None) -> Dict[str
...
                 copy_to.append(prop.type.group)
-            prop_mapping[name] = make_type_field(
-                prop.type,
-                copy_to=copy_to,
-                existing=prop_mapping.get(name, None),
+            prop_name_to_fields[name].append(
+                make_type_field(prop.type, copy_to=copy_to)
             )
 
+    prop_mapping: Dict[str, MappingProperty] = {}
```
jbothma, line 100:
```suggestion

    # Resolve list of field definitions to a single definition per field name
    prop_mapping: Dict[str, MappingProperty] = {}
```

yente/search/mapping.py
```diff
@@ -56,16 +61,10 @@ def make_field(
...
-    existing: Optional[MappingProperty] = None,
 ) -> MappingProperty:
     field_type = "keyword" if type_.group else "text"
     if type_ in TEXT_TYPES:
         field_type = "text"
-
-    # keyword type trumps text type for conflicting property types
-    # e.g. UnknownLink/Email:subject
-    if existing and existing["type"] == "keyword" and field_type == "text":
-        return existing
```
jbothma, line 68: nice cleanup

tests/test_mappings.py
```diff
@@ -0,0 +1,101 @@
...
+    finally:
+        # Clean up the test index
+        await search_provider.delete_index(temp_index)
+
+
+def test_colliding_prop_names():
+    """Test that we can handle multiple properties with the same property name."""
+    mapping = make_entity_mapping()
+    # Yo dawg, I heard you like properties
+    prop_mapping = mapping["properties"]["properties"]["properties"]
```
jbothma, line 98: really?!
(not a review comment, just exasperation)

tests/test_mappings.py
```diff
@@ -0,0 +1,101 @@
...
+    # CallForTenders:authority is an entity, Identification:authority is a string
+    assert set(prop_mapping["authority"]["copy_to"]) == set(["text", "entities"])
```
jbothma, line 100: nice and clear

tests/test_mappings.py
jbothma: really.

tests/test_mappings.py
```diff
@@ -0,0 +1,101 @@
...
+        )
+        assert len(search_result["hits"]["hits"]) == 1, "Failed to match on names"
+
+        # name_parts and name_phonetic are a bit of a special case, we syntesize them in the indexer
+        search_result = await search_provider.search(
+            temp_index, {"bool": {"must": [{"match": {"name_parts": "Vladimir"}}]}}
+        )
+        assert len(search_result["hits"]["hits"]) == 1, "Failed to match on name_parts"
+        search_result = await search_provider.search(
+            temp_index, {"bool": {"must": [{"match": {"name_phonetic": "FLTMR"}}]}}
```
jbothma, line 62: sounds like you're whispering

jbothma (approved):
this is a solid PR. Very nice. The tests look good. I've added a couple of thoughts but no hard requests.

I skipped the benchmark stuff. Not sure if that's supposed to be part of this?

yente/search/mapping.py
jbothma: only "# Collect field definitions:" is new

yente/search/mapping.py
leonhandreke: Yeah it's dumb.

PR #848, opened 2025-08-18
Improved retry strategy in httpx transport

rae-306:
Ingestion failed for us because of our bumpy network.

It turns out that the retry strategy in `stream_http_lines` is quite basic: retry 3 times and sleep 1 second in between.

This PR improves the `stream_http_lines` function retry strategy. It does so by defining a global retry strategy (exponential backoff) at the level of the `httpx_session` client defined in yente.data.util.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,7 @@ dependencies = [
     "opensearch-py[async] == 2.8.0",
     "uvicorn[standard] == 0.35.0",
     "httpx[http2] == 0.28.1",
+    "httpx-retries",
     "aiohttp[speedups] == 3.12.15",
     "boto3 ~= 1.37",
     "fastapi == 0.116.1",


rae-306:
Hi Leon, thanks for your response. After hours of debugging, we finally managed to get it working. Our bumpy network would cause a TransportError every now and then while we were loading the "default" dataset with 4.1M entries.
There were mainly two issues we discovered in the code, for which I intend to submit a PR. Perhaps it is better to close this one and begin anew. Anyhow, the two bugs are:

1) The `stream_http_lines` should retry 3 times. However, if the `client.stream()` throws a TransportError the first time, but works correctly the second time, then the retry counter is not reset. Which means that if the TransportError is thrown more than three times in total, then the retry fails completely. 

2) As you mention, there is no "range-request" magic in the dataset JSON. So if a TransportError is thrown, then the stream starts from the beginning of the JSON. Of course, the process is idempotent: the doc count in ES stays constant and until it catches up with where it left, it simply immediately deletes every new file it inserts. However, there are two problems with this:
a) Your counter from `yente.search.indexer` "Index: ... entities..." is not reset / continues to grow.
b) In our case, if a network is bumpy enough, then this process would go on and on. Suppose that every few minutes there is a small hiccup in the network, then the loading of the full json would take forever.
What we did was we changed the `stream_http_lines` function to do the following. We make the function keep track of the "last_id" and on a retry it reads the json until it find the last_id and only yields from that point on. Since reading a json is way faster than inserting in Elasticsearch, this speeds up enormously. Also the counter now gives the correct amount :)

Issue #849, opened 2025-08-20
Confirm copy_to merges on `existing` mappings


pudo:
I think we need to be careful here: https://github.com/opensanctions/yente/blob/9a44e261de53189a863353b053893b7d6907b711/yente/search/mapping.py#L67-L69 - I think the `copy_to` of the existing and the new field should be merged.

leonhandreke:
See the commit above. It seems the only collision left with FtM 4.0 is `authority` (sometimes it's string, sometimes entity), and since all keyword fields are also copied to `text`, the merging is a de-facto no-op. But at least we do it properly now. Let's hope we don't get more scary type collisions in the future.

Issue #822, opened 2025-08-04
Explore: indexing symbols to improve matcher recall


pudo:
We now have a `rigour.names.tagging` module that allows us to assign symbols to the names of companies and people. For example, you could tag the QIDs of name parts like this: 

```
>>> from rigour.names import Name, tag_person_name
>>> from rigour.names.tokenize import normalize_name
>>> name = Name("John Smith")
>>> tag_person_name(name, normalize_name).symbols
{<Symbol(Category.NAME, 104552334)>, <Symbol(Category.NAME, 4925477)>, <Symbol(Category.NAME, 1158446)>}
```

Similarly for companies:
```
>>> from rigour.names import tag_org_name
>>> name = Name("OOO Saturn Holding")
>>> tag_org_name(name, normalize_name).symbols
{<Symbol(Category.ORG_CLASS, LLC)>, <Symbol(Category.SYMBOL, HOLDING)>}
>>> name = Name("OOO Saturn Holding")
>>> tag_org_name(name, normalize_name).symbols
{<Symbol(Category.ORG_CLASS, LLC)>, <Symbol(Category.SYMBOL, HOLDING)>}
```

The point here is that these symbols buffer out language to some extent: 

```
>>> name = Name("OOO Saturn холдинг")
>>> tag_org_name(name, normalize_name).symbols
{<Symbol(Category.ORG_CLASS, LLC)>, <Symbol(Category.SYMBOL, HOLDING)>}
>>> name = Name("John смит")
>>> tag_person_name(name, normalize_name).symbols
{<Symbol(Category.NAME, 104552334)>, <Symbol(Category.NAME, 27106112)>, <Symbol(Category.NAME, 4925477)>, <Symbol(Category.NAME, 1158446)>}
```

What we need to understand is if we want these to be indexed in ElasticSearch in order to improve recall, or if the additional memory and time used for doing this outweights the benefits.

simonwoerpel:
wow this is super interesting for the new OpenAleph index as well. i can see this as a huge improvement to the current synonyms implementation and as well for matching. is there already a PR or some working on this i can look into (or even contribute :))?

leonhandreke:
This is done and released now and seems to work okay.

PR #798, opened 2025-07-16
Port to FtM/nomenklatura 4

pudo:

Issue #748, opened 2025-05-16
Implement a mutex in ES to avoid concurrent indexes


pudo:
We keep running into scenarios where an index job gets triggered multiple times, e.g. when our update frequency in GCP is lower than the time it takes to index. I wonder if we should take the hit and put a special index, `{prefix}-lock` where we insert a document with the index name and then delete it at the end of the job? The mutex would likely include a timestamp so that "lost jobs" can be detected after a timeout. 

Alternatively, we could frame this as an "indexer log", which stores all index runs and their status and doesn't delete the run info after the fact. This may be fun to have for audit trail purposes.

leonhandreke:
Just thought of an alternative: maybe we could use the last modified time of an index and just not touch it if it's been written to in the last 30s or so? See https://stackoverflow.com/questions/32584581/how-to-get-the-last-update-of-the-index-in-elasticsearch for how to get that number.

pudo:
Oh I had no idea that indexes had such metadata. That's really cool, and yes: it should actually address the issue at hand. 

The counter-thought is that having an audit log of index updates (when was an index started, completed, etc. - when did which dataset become available in screening) might be a whole feature onto itself.

jbothma:
We're setting out to solve

- reindex job interval is shorter than the time it takes to do a full re-index or apply a delta
- misconfiguration leads to concurrent update events, e.g. api workers don't have the threaded index disabled

We're not setting out to solve

- two instances of yente with different namespaces concurrently indexing in the same elasticsearch cluster
- perfection - index jobs are considered idempotent, and while concurrent jobs might mean a performance hit, rare corner cases are considered tolerable

PR #864, opened 2025-08-28
leonhandreke/index subversion

leonhandreke:
- **Port to FtM/nomenklatura 4 (#798)**
  * Begin to port to ftm4/nk4
  
  * Simplify type mess a bit
  
  * bump versions
  
  * resolve path
  
  * womp womp
  
  * womp womp, the musical
  
  * Bring back Entity.target
  
  * add types-PyYAML to dev deps
  
  ---------
  
  Co-authored-by: Leon Handreke <leonh@ndreke.de>

- **Bump dependencies a bit once more**
  

- **Fix /algorithms broken by new nomenklatura**
  

- **Fix more nomenklatura 4.x breakage**
  

- **Bump cryptography from 45.0.5 to 45.0.6**
  Bumps [cryptography](https://github.com/pyca/cryptography) from 45.0.5 to 45.0.6.
  - [Changelog](https://github.com/pyca/cryptography/blob/main/CHANGELOG.rst)
  - [Commits](https://github.com/pyca/cryptography/compare/45.0.5...45.0.6)
  
  ---
  updated-dependencies:
  - dependency-name: cryptography
    dependency-version: 45.0.6
    dependency-type: direct:production
    update-type: version-update:semver-patch
  ...
  
  Signed-off-by: dependabot[bot] <support@github.com>
  

- **Warm up matcher algorithms at startup**
  From https://fastapi.tiangolo.com/advanced/events/:
  
  > This code will be executed before the application starts taking requests, during the startup.
  
  https://github.com/opensanctions/yente/pull/831
  

- **Adjust function signature and docs of SearchProvider.bulk_index**
  

- **Add YENTE_HIDDEN_ALGORITHMS env var to hide logic-v2 from prod for now (#832)**
  * Add YENTE_HIDDEN_ALGORITHMS env var to hide logic-v2 from prod for now
  
  Question: Is the `assert` in `settings` a bit over the top? Should it be
  an error/warning log in `lifespan` instead?
  
  * Create a yente-internal enabled algorithms list containing logic-v2
  
  Algorithm selection is now case-sensitive
  
  * Don't change behavior for get_algorithm
  
  * Move startup check to routers/util.py instead of settings

- **Implement passing scoring config in match request**
  

- **Fix algorithm selection from passed param**
  

- **Fix test broken by hiding logic-v2**
  

- **Reply with a docs field in /algorithms, deprecate features**
  

- **Remove unused pantomime**
  

- **Try to speed up builds**
  

- **Use rigour for env var handling**
  

- **Vibecode a little candidate generation benchmark tool**
  

- **Allow passing mappings and settings into SearchProvider.create_index**
  

- **candidate_generation_benchmark: support comparing git refs**
  

- **candidate_generation_benchmark: Add positives test dataset from un_sc**
  

- **candidate_generation_benchmark: Add UN untreated dataset, improve benchmark output**
  

- **Index Entity.to_dict, rely on copy_to magic, add tests for that**
  Previously, the NAME_{KEY,PART,PHONETIC} fields were built from the
  `names` type field generated by `to_full_dict`. That's no longer being
  called, so we're now pulling out the names using
  `Entity.get_type_values` in `build_indexable_entity_doc` to do further
  shenanigans on the names.
  
  See https://github.com/opensanctions/yente/issues/806
  

- **mapping: merge copy_to fields for colliding prop names**
  Currently, the only prop that this happens for in FtM is `authority`. `CallForTenders:authority` is an entity, `Identification:authority` is a string
  
  Fixes https://github.com/opensanctions/yente/issues/849
  

- **Apply suggestion from @jbothma**
  Co-authored-by: JD Bothma <jbothma@gmail.com>

- **Remove TODO after review**
  

- **Index and match on name symbols as tagged by rigour**
  https://github.com/opensanctions/yente/issues/822
  

- **Expose matcher explanations in /match result**
  See https://github.com/opensanctions/yente/issues/821
  

- **ES index locking using an audit log (#836)**
  Log reindex operations to an audit log in ES and use that as a (theoretically non-perfect) locking mechanism to prevent concurrent reindexes.
  
  See https://github.com/opensanctions/yente/issues/748
  
  Co-authored-by: JD Bothma <jbothma@gmail.com>

- **Bump nomenklatura and rigour to pull some fixes**
  

- **[contrib/candidate_generation_benchmark] Add us_congress fixture**
  

- **Mark `matcher` field from `/match` response deprecated**
  

- **Google Cloud Logging compatible request logging**
  Google Cloud Logging has a magic httpRequest field in its LogEntry proto
  that triggers special display logic (and more?). For all the people who
  don't use Google Cloud, this isn't worse than what we had before. In
  fact, it's better, case now we can process X-Fowarded-For.
  

- **middleware: Google Cloud Logging compatible trace logging**
  

- **[middleware] Refactor TraceContext a bit with @dataclass**
  

- **search: dial down length normalization on names & up-score large docs in /search (#779)**
  * search: dial down length normalization on names & up-score large docs
  
  Brought to you by the Friday-afternoon labs of search engine
  engineering! The end result here is that Vlad is first for the query
  "Putin", and that Daniel Putin, the former Mayor of Champagnat, is a bit
  further down.
  
  Length normalization on names is a bit of a difficult topic for us:
  Important entities for us usually have a lot of names, and if only some
  of them match, we shouldn't downrank it too much for having lots of others.
  This is the reason why currently, small docs with one exact name match
  always come up first. However, we also have to be careful not to rank
  "Putin’s chef" too highly for the "Putin" query, so we don't dial it all
  the way down to 0.
  
  This commit also introduces a num_values field in the index that we use
  to up-score "important" docs a bit.
  
  * Only upscore large entities in /search

- **audit_log: Fix bash one-liner for retrieval**
  

- **Introduce an IndexInfo dataclass instead of tuples**
  

- **Restructure index name building a bit**
  

- **Remove weird arg match_system_version to parse_index_name**
  Pass a bool to throw a ValueError, that's just weird. Do it the elegant
  functional way.
  

- **Split index version by - instead of relying on system version length**
  

- **Rename system_version -> get_system_version**
  

- **Introduce YENTE_INDEX_SUBVERSION env to force a full reindex**
  

- **Add a little index name parsing test**


leonhandreke:
Can we think of a better name?

leonhandreke:
Haha no, I think `rcs` is the newest one I haven't used at some point in my life :)

I like `YENTE_INDEX_REBUILD_ID`!

PR #866, opened 2025-08-28
Log detail about the coordinating node and serving shards in debug mode

jbothma:
This has a performance impact because it enables explain and counting total possible results from elasticsearch.

Do we want it to be a different option than YENTE_DEBUG?

### DIFF
--- a/yente/cli.py
+++ b/yente/cli.py
@@ -33,14 +33,14 @@ def serve() -> None:
             server_header=False,
         ),
     )
-    configure_logging()
+    configure_logging(settings.LOG_LEVEL)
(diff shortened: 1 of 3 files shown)


pudo:
Please let's make sure we don't log response contents in the application - I would prefer not to even have it in a debug mode so it doesn't get activated by accident.

jbothma:
Good point - let's close this. This is useful to understand the candidates returned for a given query but I think a lot of its utility is in logging the IDs, but salting and hashing that for an extremely occasionally-used log entry is overkill at the moment.

PR #867, opened 2025-08-28
WIP idea: Filter out org class parts from names query

leonhandreke:


leonhandreke:
https://github.com/opensanctions/yente/pull/869

PR #868, opened 2025-08-28
WIP Idea: For phonemes and name parts, use a max instead of a sum of parts

leonhandreke:
Hoping that will kick out the triple-scoring ot "LTD". First testing indicates it does somewhat, but if the phoneme is very short (like "Oil"), we end up using the score for LTD again.


leonhandreke:
https://github.com/opensanctions/yente/pull/869

PR #870, opened 2025-09-01
Let UPDATE_TOKEN not have an unsafe-default, default to disabled

leonhandreke:


leonhandreke:
@pudo maybe there is a reason you did it like that?

yente/settings.py
```diff
@@ -137,7 +137,7 @@ def random_cron() -> str:
 BASE_SCHEMA = "Thing"
 PORT = int(env_str("YENTE_PORT", env_str("PORT", "8000")))
 HOST = env_str("YENTE_HOST", env_str("HOST", "0.0.0.0"))
-UPDATE_TOKEN = env_str("YENTE_UPDATE_TOKEN", "unsafe-default")
+UPDATE_TOKEN = env_str("YENTE_UPDATE_TOKEN", "") or None
```
pudo, line 140: I think there's an `env_opt` that is `Optional[str]`.

pudo (approved):
No I think the original "thinking" was making it easy to deploy, but in the end it just feels awful. Let's apply this patch.

PR #871, opened 2025-09-01
docs: Move upgrading from FAQ to yente docs

leonhandreke:
- **docs: Port how to upgrade FAQ item**
  

- **indexer: Comment to clarify the implications of create_index**
  

- **docs: Don't force people to re-build the index on upgrade**

### DIFF
--- a/docs/deploy/upgrading.md
+++ b/docs/deploy/upgrading.md
@@ -22,13 +22,13 @@ The following instructions document the commands when running with Docker Compos
 4. Restart the yente container: `docker compose restart app`
 
 
-## Can I deploy a new version of yente incurring little to no downtime?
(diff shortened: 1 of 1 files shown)


leonhandreke:
@pudo I reworded the docs a bit from the FAQ item. Some questions:

- Are we OK stating that we best-effort try to ensure that new yentes can read old indices for low-downtime upgrades? We rely on this ourselves, so seems like we can just state it (and also drop the requirement for `yente reindex -f` on upgrade).
- Remove "check out new scoring system" - I feel like this is quite specific, doesn't happen so often and will also be prominently featured in the changelog? My idea to get rid of one bullet point to make the list of things to do seem less involved.

PR #876, opened 2025-09-03
Put system_version in index alias name to allow for blue-green rollouts

leonhandreke:
- **Put system_version in index alias name to allow for blue-green rollouts**
  This effectively makes the system version (i.e. the yente index version as well as the FtM version) part of the namespace that yente is operating and touching things in. Things outside the namespace should not be touched at all.
  
  The idea would be for a new `yente reindex` to run, set up the new alias, while the old version can keep serving from the old alias.


leonhandreke:
See also our internal ticket https://github.com/opensanctions/operations/issues/1158 where we discuss how to do canary rollouts of our internal instance of yente.

PR #879, opened 2025-09-05
reconcile: Score fewer candidates, dial down batch size

leonhandreke:
- **Use limit_window correctly in reconcile to not score 500 by default**
  Previously, `limit_window` would not only accept an int, but also accept a `None` and then replace that with a default if it didn't parse as an int. Most of the callers never used that, so dumb it down a bit and make the default explicit in `/reconcile`, which is the only place where that default setting was used, and it was used wrong, leading to 500 candidates being scored every time (I think that's what was happening at least)
  

- **reconcile: Dial batch size down from 10 to 5**
  Previously, it was set to settings.DEFAULT_PAGE (for whatever reason, the default number of search results shouldn't really have any connection with the batch size). Make the batch size a bit smaller to make the requests to our API more equal in weight.

### DIFF
--- a/yente/routers/reconcile.py
+++ b/yente/routers/reconcile.py
@@ -99,7 +99,7 @@ async def reconcile(
         schemaSpace=typed_url("https://www.opensanctions.org/reference/#schema"),
         view=FreebaseManifestView(url="https://www.opensanctions.org/entities/{{id}}/"),
         documentation=typed_url("https://www.opensanctions.org/docs/"),
-        batchSize=settings.DEFAULT_PAGE,
+        batchSize=3,
         preview=FreebaseManifestPreview(
             url="https://www.opensanctions.org/entities/preview/{{id}}/",
             width=430,


yente/routers/reconcile.py
```diff
@@ -99,7 +99,7 @@ async def reconcile(
         schemaSpace=typed_url("https://www.opensanctions.org/reference/#schema"),
         view=FreebaseManifestView(url="https://www.opensanctions.org/entities/{{id}}/"),
         documentation=typed_url("https://www.opensanctions.org/docs/"),
-        batchSize=settings.DEFAULT_PAGE,
+        batchSize=5,
```
pudo, line 102: 3?

pudo (approved):
I'm reasonably proud of myself for just how confused this was before you came in. Thanks for debugging!

leonhandreke:
Actually, let's just merge it :)

Issue #542, opened 2024-10-18
Opensearch IAM auth


kizmanj:
Already posted about this in another ticket, but that issue has been closed since.

Tried using IAM auth to connect to OpenSearch from a container running in EKS, with 
https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html

Yente 4.1.0 seems to have updated components that helped, but did not fully resolve the problem.

Using these settings:
export set YENTE_INDEX_TYPE="opensearch"
export set YENTE_INDEX_URL="https://redacted/"
export set YENTE_OPENSEARCH_REGION="eu-central-1"
export set YENTE_OPENSEARCH_SERVICE="es"

The error currently looks like this:

Traceback (most recent call last):
File "/venv/bin/yente", line 33, in
sys.exit(load_entry_point('yente', 'console_scripts', 'yente')())
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/venv/lib/python3.12/site-packages/click/core.py", line 1157, in call
return self.main(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/venv/lib/python3.12/site-packages/click/core.py", line 1078, in main
rv = self.invoke(ctx)
^^^^^^^^^^^^^^^^
File "/venv/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
return _process_result(sub_ctx.command.invoke(sub_ctx))
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/venv/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
return ctx.invoke(self.callback, **ctx.params)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/venv/lib/python3.12/site-packages/click/core.py", line 783, in invoke
return __callback(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/app/yente/cli.py", line 44, in reindex
asyncio.run(update_index(force=force))
File "/usr/lib/python3.12/asyncio/runners.py", line 194, in run
return runner.run(main)
^^^^^^^^^^^^^^^^
File "/usr/lib/python3.12/asyncio/runners.py", line 118, in run
return self._loop.run_until_complete(task)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #881, opened 2025-09-09
Use AsyncHTTPConnection in OpenSearch to make AWS IAM work

pudo:
Fixes https://github.com/opensanctions/yente/issues/542


leonhandreke:
That should fix the tests. I don't have access to an IAM-auth OpenSearch right now, but will hopefully get to verify this soon.

PR #882, opened 2025-09-09
Introduce simple per-prefix locking mechanism

leonhandreke:
- **Simple locking mechanism that lock the whole settings.INDEX_NAME prefix**
  This is a slightly modified version of https://github.com/opensanctions/yente/pull/835. The lock document's name is no longer the index we're locking on, but just `lock`, meaning it's just a global lock that applies to all yentes sharing a  `YENTE_INDEX_NAME`.
  

- **lock: Introduce a session id to better detect and debug issues**
  

- **indexer: Also clean up lock if indexing fails (or Ctrl-C in testing)**

### DIFF
--- a/yente/search/indexer.py
+++ b/yente/search/indexer.py
@@ -200,7 +200,7 @@ async def index_entities(
             async for item in it:
                 idx += 1
                 # Refresh the lock every 50,000 documents. Should be enough not to
-                # lose the lock, expiration time is lock.LOCK_EXPIRATION_TIME (currently 5 minutes)
+                # lose the lock, expiration time is lock.LOCK_EXPIRATION_TIME (currently 10 minutes)
                 if idx % 50000 == 0:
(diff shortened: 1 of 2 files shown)


yente/search/indexer.py
```diff
@@ -314,12 +284,24 @@ async def update_index(force: bool = False) -> None:
...
+            log.warning("Failed to acquire lock, skipping index update")
+            return
+        try:
+            for dataset in catalog.datasets:
+                with lock:
+                    await index_entities(
+                        provider, dataset, force=force, lock_session=lock_session
+                    )
+
+            await delete_old_indices(provider, catalog)
```
pudo, line 298: Yeah this feels nice.

pudo (approved):
Ok I feel good about this. It's been thought about a fair bit by @jbothma and you. Do you want to put it in `test` for the morning?

Issue #902, opened 2025-09-30
Multiple catalogs with dataset name conflicts has undefined behaviour


jbothma:
Two catalogs with overlapping dataset names can have surprising behaviour. e.g. any datasets loaded by the second catalog that aren't the direct scope of the first catalog would be deleted because catalog.get(dataset.name) matches the first in order of loading datasets.

jbothma:
This doesn't seem to be new behaviour caused by indexer changes, e.g. https://github.com/opensanctions/yente/pull/845/files?diff=split&w=0#diff-be09cb6136eba10e3c8a46e4c804b0b452383f045c13988a7e690f5fe47041edR281

It might be due to catalog changes but I suspect this has been there a long time. Still worth 
- making the manifest load process 
  - skip datasets added by earlier catalogs
  - warn that they're being skipped and that it's a configuration problem
- clarifying the documentation

jbothma:
A different but related issue: If a scope isn't found in the catalog, I think it silently just doesn't index anything from that catalog entry in the manifest. 

That's noticable if it's your only catalog, but maybe it could warn so that someone who doesn't know what indexing logs to expect can more easily understand the problem.

pudo:
Yes it was.

PR #923, opened 2025-10-21
Raise an exception if scope not found in defined catalog

leonhandreke:
https://github.com/opensanctions/yente/issues/902


leonhandreke:
@jbothma just a second pair of eyes, did you have any reason to suggest in the bug to merely warn? I feel like erroring out is appropriate for supplying an invalid scope, you might be missing things that you expected to be there. I'm thinking maybe people specifying a dataset and us renaming it or something like that... WDYT?

pudo:
Agree with this being a proper exception - nice catch!

PR #945, opened 2025-11-05
Wait for index before starting app in docker compose

jbothma:
This should prevent errors at startup that are often a bit confusing and not obviously ignorable.

Thanks @mainibet

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,7 +20,8 @@ services:
     volumes:
       - index-os-data:/usr/share/elasticsearch/data
     healthcheck:
-      test: ["CMD-SHELL", "curl -f http://localhost:9200/_cluster/health || exit 1"]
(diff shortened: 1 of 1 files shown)


pudo:
I'm so confused by why we haven't had this until now :) Very nice fix. The only thing to check perhaps is since when `condition_healthy` has been in `docker compose` and document that as a new minimum version.

leonhandreke:
Re: version compatibility: `condition: service_healthy` was added in May of 2020 https://github.com/compose-spec/compose-spec/commit/cdeb25440a673100e2192213caecb3c6053cf75d. I find it really hard to figure out what version that was, they don't seem to tag their releases or make past versions of the spec available. But it's been around for a while :)

leonhandreke:
I think this is a good change, but we should definitely announce this in the release notes, just in case people are using the vanilla docker-compose file.

PR #966, opened 2025-12-03
match: Log number of candidates returned from ES

leonhandreke:


pudo:
I think it's fair to call this telemetry and not payload, so I'm +1 on logging it. Wondering if it might be easier to just log the number of candidates upon receipt of response rather than passing it around ... and if we pass it around: can we call it `candidates`, not `matches`?

PR #995, opened 2026-01-07
doc(index.md):fix double slash in hosted API link

mainibet:
Fix: removed the extra slash in hosted API link that was giving this error:
image

**Before**: `{{ config.extra.opensanctions_url }}//api/` → `https://www.opensanctions.org//api/`

**After**: `{{ config.extra.opensanctions_url }}/api/` → `https://www.opensanctions.org/api/`


jbothma:
Thank you!

PR #1082, opened 2026-03-31
contrib: add validation_report tool

leonhandreke:
Adds `contrib/validation_report/` — a small tool to validate match quality against known-good fixture CSVs by calling a running yente instance over HTTP (rather than importing internals). Tests logic-v1 vs logic-v2 side-by-side.

Two scripts:
- `generate_report_data.py` — POSTs fixture records to `/match/{dataset}` in batches and writes raw `{score, match}` results to JSON
- `render_report_data.py` — reads the JSON, computes stats (mean top score, matches, %) and renders an HTML report via Jinja2 + markdown-it-py with GitHub Markdown CSS

Fixtures are copied from `candidate_generation_benchmark` (truncated to 1000 records each).

Example: [report.html](https://github.com/user-attachments/files/26381708/report.html)

Related: https://github.com/opensanctions/operations/issues/2203


contrib/validation_report/templates/report.md.j2
```diff
@@ -0,0 +1,31 @@
...
+
+| Algorithm | Mean Top Score | Matches | Total | % Matches |
+|-----------|---------------|---------|-------|-----------|
+{% for algo, stats in algorithms.items() -%}
+| {{ algo }} | {{ "%.3f"|format(stats.mean_top_score) }} | {{ stats.matches }} | {{ stats.total }} | {{ "%.1f"|format(stats.pct_matches) }}% |
+{% endfor %}
+{% endfor %}
+## Environment
+
+**yente:** `{{ yente_git_version }}`
```
jbothma, line 40:
```suggestion
**yente git hash:** `{{ yente_git_version }}`
```

contrib/validation_report/generate_report_data.py
```diff
@@ -0,0 +1,143 @@
...
+        )
+        resp.raise_for_status()
+        data = resp.json()
+        for key in queries:
+            hits = data["responses"][key].get("results", [])
+            if hits:
+                top = hits[0]
+                results.append({"score": top["score"], "match": top["match"]})
+            else:
+                results.append({"score": 0.0, "match": False})
```
jbothma: should the positives include entity IDs and confirm that precisely the entity we're expecting was one of the results considered a match? 

If we're merely considering _some_ "match" as a match, maybe that could be a note in the report somewhere under "methodology" or something so that code doesn't have to be read to discover that.

jbothma (approved):
I think all of these thoughts are bonus - this is already very informative, but I think we can be more precise and reproducible. Feel free to break out into a new ticket or tell me why it's not applicable: 
 
- at some point the fixtures will be inconsistent with what's loaded in yente, right?
  - let's document how to update the fixtures, or how to load precisely the version of data that corresponds with the fixtures
  - Can this capture the versions of the relevant datasets actually loaded in the yente under test?
- it'd be nice if the yente readme documented how to run this

It's nice that this is in the yente repo so people can reproduce it.

contrib/validation_report/templates/report.md.j2
leonhandreke: It will be a tag if we're testing a clean tag, so "git version" will be better.

Issue #1011, opened 2026-02-02
Heuristics to score fewer index candidates


pudo:
So I've been doing a lot of perf benchmarking over the weekend on `logic_v2`, and one take-away from that is: it's not actually that slow. Yente is just using it a whole lot. We have `YENTE_MAX_MATCHES` set to 10 by default, and users of the `/match` API can request up to 500 results - that would mean: 5000 calls of the scoring algorithm. Even in the default case (with 5 results), we compare 50 candidates. 

This is done because we don't believe that the scoring of ES and our algorithm are well aligned. Is that actually true? To what extent? It would be a dream to have some numbers on how much they diverge on average. 

But even beyond that: do we always need to score all results? Imagine a result set with 50 candidates, but after the 3 highest-ranking have been scored, none of them have reached `cutoff`, or even `cutoff/2`. Do we want to keep going? One approach might be for the scoring function to collect statistics by index and - once it has seen a sufficient number of samples - begin to dynamically set a conservative cutoff value.

pudo:
## Research: production scoring log analysis

Analyzed three production log samples (30,000 rows, ~20,800 valid scoring entries, 418 unique queries) to understand scoring behavior and evaluate early stopping strategies.

### Most scoring work is wasted

| Metric | Value |
|---|---|
| Total scoring calls | 20,772 |
| Scores < 0.5 (below cutoff) | 82.2% |
| Scores < 0.3 (clearly wasted) | 47.9% |
| Scores >= 0.7 (match threshold) | 1.0% |
| Queries with zero candidates >= 0.5 | 49.3% |
| Queries with zero candidates >= 0.7 | 84.4% |

About half of all queries produce no candidates above 0.5, and 84% produce no matches (>= 0.7). Yet we score all ~50 candidates for every query.

### ES ranking vs algo score correlation

ES ranking is a **weak** predictor of algo score. The best algo-scored result appears at:

| Within top N ES results | % of queries |
|---|---|
| Top 1 | 23.2% |
| Top 5 | 44.7% |
| Top 10 | 63.4% |
| Top 20 | 83.3% |
| Top 50 | 98.3% |

Within the first 50 candidates, algo scores decline gently (mean 0.36 → 0.30) but good results appear at every rank bucket. ES excludes truly irrelevant candidates past rank 50 (0% with algo >= 0.5 there), but within the top 50 it cannot reliably distinguish good from bad.

[...]

PR #1096, opened 2026-04-07
Early stopping via score budget for candidate scoring

pudo:
## Summary

- Add early stopping to `score_results()` using a score budget: each candidate costs 1 token, and its score earns back `score / (threshold/2)` tokens. When the budget is exhausted, stop scoring.
- One env var: `YENTE_SCORE_EARLY_STOP_BUDGET` (default 10, set high to disable)
- All other values derived from the per-request `threshold` parameter
- On production data (418 queries): saves ~27% of scoring calls, misses 3 results (0.7%), all sub-threshold

The formula:
```
budget = budget - 1 + score / (threshold / 2)
stop when budget <= 0 and rank >= limit
```

A score of `threshold/2` (0.35 at default) breaks even. Higher scores extend the search; lower scores drain the budget. This naturally adapts to query quality — queries with real matches keep searching proportionally longer.

Refs #1011

## Test plan

- [ ] Verify existing tests pass
- [ ] Compare `/match` output with and without early stopping on representative queries
- [ ] Monitor miss rate in production logs

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -36,5 +36,6 @@ Yente features various configuration options related to data refresh and re-inde
 | `YENTE_MAX_MATCHES` | `500` | How many results to return per `/match` query at most. |
 | `YENTE_MATCH_CANDIDATES` | `10` | How many candidates to retrieve from the search as a multiplier of the `/match` limit. Note that increasing this parameter will also increase query cost, as each of these candidates scored after retrieval from the index.|
 | `YENTE_MATCH_FUZZY` | `true` | Whether to run expensive Levenshtein queries inside ElasticSearch. |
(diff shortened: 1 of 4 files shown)


leonhandreke:
Looks very good. The only thought I had was that `YENTE_SCORE_EARLY_STOP_BUDGET` could be `limit * 2` or something? Right now, it's a bit of an environment-level constant of how many results we *actually* expect to make sense IIUC. If limit is higher and the budget is exhausted, I understand we only continue until we've filled up our `limit` results and then exit.

But yeah, that's the only original thought I had and I'm not really convinced it's a good one. But wanted to throw it out there.

leonhandreke:
@pudo any thoughts on my only human thought above? If your thought is "meh..." that's also fine, I'm happy to merge like this, I don't have a strong opinion, just wanted to make sure my original human thought didn't go to waste :P

pudo:
So it turns out that the stop budget is actually unrelated to the limit - I also didn't fully grok this. So they both happen to have a similar "order of magnitude", but it makes no logical sense to create a link between them. 

This PR still frightens me a bit. Do you think you could use the new framework to look at the before and after?

leonhandreke:
We've done a lot for performance and currently it isn't an issue anymore. Let's re-vibe it if it comes up again.

PR #1113, opened 2026-04-21
Add Subresource Integrity for ReDoc bundle

goose-life:
Requires https://github.com/opensanctions/knowledgebase/pull/22

## Summary

- Pins the ReDoc bundle to a versioned URL (`redoc-2.5.2.standalone.js`) instead of a floating path
- Injects `integrity` and `crossorigin` attributes onto the `<script>` tag so the browser verifies the SHA-384 hash before executing the file — a tampered or substituted bundle will be rejected
- Adds `YENTE_REDOC_JS_URL` and `YENTE_REDOC_JS_SRI` env vars for operators who need to override the defaults

## Upgrading ReDoc

1. Bump `REDOC_VERSION` in `knowledgebase/Makefile` and run `make assets` to upload the new bundle to `assets.opensanctions.org`
2. Update `YENTE_REDOC_JS_URL` to the new versioned path and recompute the hash:
   ```
   curl -sL <new_url> | openssl dgst -sha384 -binary | openssl base64 -A
   ```
3. Update `YENTE_REDOC_JS_SRI` with the result (prefixed `sha384-`)

## Test plan

- [x] `pytest tests/test_app.py::test_redoc_has_sri` passes without Elasticsearch
- [x] `GET /` on a running instance renders the ReDoc docs page correctly

### DIFF
--- a/yente/routers/admin.py
+++ b/yente/routers/admin.py
@@ -38,9 +38,23 @@ async def redoc_html() -> HTMLResponse:
     # script's SHA-384 hash before executing it. If the server ever delivers a different
     # file the hash won't match and the browser refuses to run it.
     # Update YENTE_REDOC_JS_URL and YENTE_REDOC_JS_SRI together when upgrading ReDoc.
-    html = bytes(response.body).decode("utf-8").replace(
-        f'src="{settings.REDOC_JS_URL}"',
-        f'src="{settings.REDOC_JS_URL}" integrity="{settings.REDOC_JS_SRI}" crossorigin="anonymous"',
+    # onerror fires on SRI mismatch or network failure (unlike <noscript>, which only
+    # fires when JavaScript is disabled entirely).
+    fallback = (
+        '<p id="redoc-error" style="display:none;padding:2em">'
+        "If you see this, you may need to update <code>YENTE_REDOC_JS_SRI</code> "
+        "to match the script at <code>YENTE_REDOC_JS_URL</code>.</p>"
+    )
+    html = (
+        bytes(response.body)
+        .decode("utf-8")
+        .replace(
+            f'src="{settings.REDOC_JS_URL}"',
+            f'src="{settings.REDOC_JS_URL}" integrity="{settings.REDOC_JS_SRI}"'
+            ' crossorigin="anonymous"'
+            ' onerror="document.getElementById(\'redoc-error\').style.display=\'block\'"',
+        )
+        .replace("<redoc ", f"{fallback}\n    <redoc ")
     )
     return HTMLResponse(html)
 
--- a/yente/routers/admin.py
+++ b/yente/routers/admin.py
@@ -25,14 +25,24 @@ router = APIRouter()
     include_in_schema=False,
 )
 async def redoc_html() -> HTMLResponse:
-    """Render the ReDoc API documentation renderer script."""
-    return get_redoc_html(
+    """Render the ReDoc API documentation page."""
+    response = get_redoc_html(
         openapi_url="/openapi.json",
         title=settings.TITLE,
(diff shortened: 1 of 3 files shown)


yente/routers/admin.py
```diff
@@ -25,14 +25,24 @@
...
     )
+    # get_redoc_html produces a plain <script src="..."> tag with no integrity check.
+    # We inject Subresource Integrity (SRI) attributes so the browser verifies the
+    # script's SHA-384 hash before executing it. If the server ever delivers a different
+    # file the hash won't match and the browser refuses to run it.
+    # Update YENTE_REDOC_JS_URL and YENTE_REDOC_JS_SRI together when upgrading ReDoc.
+    html = bytes(response.body).decode("utf-8").replace(
+        f'src="{settings.REDOC_JS_URL}"',
+        f'src="{settings.REDOC_JS_URL}" integrity="{settings.REDOC_JS_SRI}" crossorigin="anonymous"',
+    )
```
jbothma: an alternative would just be to make a template out of view-source:https://api.opensanctions.org/ and render it here. I'm not sure which is better.

One benefit of templating is that we can add a note about checking subresource integrity to the noscript tag (does that show if subresource integrity rejected the bundle?)

leonhandreke:
See https://opensanctions-corp.slack.com/archives/C066UMQDNPQ/p1776950871367429 - let's go with CDN + SRI?

goose-life:
~This would mean cutting out https://github.com/opensanctions/knowledgebase/pull/22 altogether; the Makefile in `knowledgebase` could stay for the next couple of yente versions and then go away.~ 

Nope, we're still using assets.opensanctions.org

goose-life:
Got this by changing a character in `YENTE_REDOC_JS_SRI`:
image

I also made `YENTE_REDOC_JS_URL` point to cdn.redoc.ly locally, and fixed the SRI to check: 
image

goose-life:
once we merge https://github.com/opensanctions/knowledgebase/pull/22 and run make or whatever there, we can repeat the above test locally before merging

yente/settings.py
```diff
@@ -219,3 +219,19 @@ def random_cron() -> str:
 
 # Used to pad out first_seen, last_seen on static collections
 RUN_DT = utc_now()
+
+# ReDoc API documentation bundle, served at GET /.
+# URL is versioned so the path is stable and the SRI hash doesn't need updating
+# unless you deliberately upgrade. To upgrade:
+#   1. Update REDOC_VERSION in knowledgebase/Makefile and run `make assets`
+#      hosted in assets.opensanctions.org to address security / privacy concerns
```
jbothma: Maybe best not to document the OpenSanctions operations stuff in yente the open source tool - we can document the procedure there instead.

```suggestion
#   1. Hosted in assets.opensanctions.org to address security / privacy concerns
```

yente/settings.py
```diff
@@ -219,3 +219,19 @@ def random_cron() -> str:
 
 # Used to pad out first_seen, last_seen on static collections
 RUN_DT = utc_now()
+
+# ReDoc API documentation bundle, served at GET /.
+# URL is versioned so the path is stable and the SRI hash doesn't need updating
+# unless you deliberately upgrade. To upgrade:
```
jbothma:
```suggestion
# ReDoc API documentation bundle used by API documentation page served at `/`.
# URL is versioned so the path is stable and the SRI hash doesn't need updating
# unless you deliberately upgrade. To upgrade:
```

jbothma (approved):
some suggestions but good to go

leonhandreke:
Let's ship it!

goose-life:
@leonhandreke I can only close or comment on the PR … would you do the honours when you have a minute? I've double-checked locally that this is gtg :pray:

PR #1087, opened 2026-04-02
feat(otel): add OpenTelemetry base instrumentation

dimoschi:
## Summary

Adds OpenTelemetry instrumentation to yente (ref #1079):

- **FastAPI auto-instrumentation** for HTTP metrics and traces via `FastAPIInstrumentor`
- **Manual spans on all SearchProvider operations** via `__init_subclass__`, providing consistent tracing for both Elasticsearch and OpenSearch backends
- Uses **stable OTEL semantic conventions** (`db.system.name`, `db.operation.name`)
- **No custom configuration** - no manual `TracerProvider`, no new yente env vars. Users configure everything via standard `OTEL_*` env vars and the `opentelemetry-instrument` CLI

### New dependencies
- `opentelemetry-api`
- `opentelemetry-sdk`
- `opentelemetry-instrumentation-fastapi`
- `opentelemetry-exporter-otlp-proto-http`

### Usage

To enable OTEL tracing, run yente with `opentelemetry-instrument`:

```bash
opentelemetry-instrument \
  --service_name yente \
  --traces_exporter otlp \
  --exporter_otlp_endpoint http://collector:4318 \
  yente serve
```

Or configure via env vars:
```bash
OTEL_SERVICE_NAME=yente
OTEL_TRACES_EXPORTER=otlp
OTEL_EXPORTER_OTLP_ENDPOINT=http://collector:4318
opentelemetry-instrument yente serve
```

Without `opentelemetry-instrument`, OTEL is a no-op (zero overhead).

### What this does NOT include (future PRs)
- `/metrics` Prometheus scrape endpoint
- Custom gauges for catalog data (datasets loaded/outdated, index staleness)

## Test plan
- [ ] New tests in `tests/test_otel.py` verify span creation for HTTP requests and SearchProvider operations
- [ ] Existing test suite passes without regressions (104 tests)

PR #1117, opened 2026-04-22
otel: zero-config OpenTelemetry tracing via distro bootstrap

leonhandreke:
Couldn't push to @dimoschi's #1087 directly, so opening this based on that — thanks for the groundwork!

I went a bit further in a couple of places:

- **`opentelemetry-distro[otlp]` instead of hand-picked deps** — the distro metapackage pulls in the SDK, the reference `Distro`/`Configurator`, and the OTLP exporter bundle. Less fragile than managing individual package versions.
- **FastAPI instrumentation is fully automatic** — dropped the explicit `FastAPIInstrumentor.instrument_app()` call. When yente runs under `opentelemetry-instrument`, the class-level `.instrument()` is applied globally before the app starts, so no wiring needed in `app.py`.
- **Dropped the `@traced` decorator and the `traced()` helper** — `elasticsearch-py` has emitted native OTel spans for every IO call since 8.13 (we're on 8.19.3), so the hand-rolled outer spans were just redundant noise.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,10 @@ dependencies = [
     "aiocsv == 1.4.0",
     "aiofiles == 25.1.0",
     "rich",
+    "opentelemetry-api >= 1.40.0, < 2.0.0",
+    "opentelemetry-sdk >= 1.40.0, < 2.0.0",
+    "opentelemetry-instrumentation-fastapi >= 0.61b0, < 1.0.0",
+    "opentelemetry-exporter-otlp-proto-http >= 1.40.0, < 2.0.0",
 ]
 requires-python = ">=3.11,<=3.14"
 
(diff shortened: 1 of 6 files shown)


dimoschi:
@leonhandreke thanks for taking care. One question, will tracing work also for OpenSearch with current implementation? That was the reason I added the decorator.

leonhandreke:
That's a very good point - adding it back for OpenSearch. Just out of interest, is that you prod deployment?

dimoschi:
> That's a very good point - adding it back for OpenSearch. Just out of interest, is that you prod deployment?

I really missed your comment as I was overwhelmed. Our production setup uses AWS OpenSearch and works good, after shipping v5.3

leonhandreke:
> > That's a very good point - adding it back for OpenSearch. Just out of interest, is that you prod deployment?
> 
> I really missed your comment as I was overwhelmed. Our production setup uses AWS OpenSearch and works good, after shipping v5.3

Cool, that's really good to know! Now I know who to ask for OpenSearch questions :)

PR #1129, opened 2026-04-29
Add indexed_dataset_version_time OTel gauge

leonhandreke:
Adds an OpenTelemetry gauge `indexed_dataset_version_time` per dataset, exposing the dataset's `last_export` as a Unix timestamp. The value is written into the index's `_meta` mapping at index time and read back from there after each successful catalog refresh — so the metric reflects the actual export time of the data sitting in ES.

A small `SearchProvider` abstraction (`set_index_metadata` / `get_index_metadata`) handles the `_meta` read/write, with replace-not-merge semantics. Cloned (incremental-reindex) indexes inherit `_meta` from the source, so we overwrite unconditionally after clone.

### DIFF
--- a/yente/app.py
+++ b/yente/app.py
@@ -18,7 +18,8 @@ from yente.data import refresh_catalog
 from yente.data.entity import Entity
 from yente.routers.util import ENABLED_ALGORITHMS
 from yente.search.indexer import update_index_threaded
-from yente.provider import close_provider
+from yente.data.metrics import update_metrics
+from yente.provider import with_provider, close_provider
 from yente.middleware import (
     MaxURLLengthMiddleware,
     RequestLogMiddleware,
@@ -29,10 +30,15 @@ log = get_logger("yente")
 ExceptionHandler = Callable[[Request, Any], Coroutine[Any, Any, Response]]
 
 
-async def cron_task() -> None:
+async def refresh_catalog_cron_task() -> None:
     await refresh_catalog()
     if settings.AUTO_REINDEX:
         update_index_threaded()
+
+
+async def update_metrics_task() -> None:
+    async with with_provider() as provider:
+        await update_metrics(provider)
 
 
 async def warm_up() -> None:
@@ -41,6 +47,7 @@ async def warm_up() -> None:
     await refresh_catalog()
     if settings.AUTO_REINDEX:
         update_index_threaded()
+    await update_metrics_task()
 
     log.debug("Warming up matcher algorithms...")
     # This is the pragmatic and easy way to warm up the matcher algorithms. If anyone feels the call to
@@ -78,7 +85,9 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
         crontab=settings.CRONTAB,
         auto_reindex=settings.AUTO_REINDEX,
     )
-    settings.CRON = aiocron.crontab(settings.CRONTAB, func=cron_task)
+    settings.CRON = aiocron.crontab(settings.CRONTAB, func=refresh_catalog_cron_task)
+    # Local handle keeps the cron alive for the lifetime of the lifespan context.
+    _metrics_cron = aiocron.crontab("* * * * *", func=update_metrics_task)
     yield
     await close_provider()
 
--- a/yente/search/indexer.py
+++ b/yente/search/indexer.py
@@ -213,6 +213,17 @@ async def index_entities(
             next_index, mappings=make_entity_mapping(), settings=INDEX_SETTINGS
         )
 
+    # Always overwrite metadata: a fresh index has none, and a cloned index
+    # inherits the source's _meta (mappings can't be overridden in a clone request).
+    #
+    # Dates in ES documents would normally be stored as timestamps, but index
+    # metadata is just an opaque blob — ES doesn't parse or process it. For
+    # consistency we store values here exactly as we got them from the catalog.
+    index_metadata: Dict[str, Any] = {}
+    if dataset.model.last_export is not None:
+        index_metadata["last_export"] = dataset.model.last_export.isoformat()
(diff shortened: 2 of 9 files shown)


leonhandreke:
@pudo how much do you hate this mechanism and do you have a better idea? Basically, I want to have a metric that makes it possible to alert on "default older than X". Right now, we can only export stale/not stale (`index_version == catalog.version`). That's another option, to just alert on "stale longer than X hours", but not as elegant. WDYT?

leonhandreke:
@pudo ping, I think this would make a nice addition to our monitoring story. The mechanism is a bit weird, see PR description first sentence. Would love to hear if you have a better idea :)

leonhandreke:
This might be a better place to put it: https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/mapping-meta-field

pudo:
Uuuh that's a really nice discovery

leonhandreke:
@jbothma adding you just for general thoughts on this - not expecting you to review every line. Just... I don't know, it's adding a thing to the index, and I feel I want a second pair of eyes for that. 

Things I can think might be interesting:
- is the periodic mechanism to update the gauge a good idea? Better ways?
- other metadata to store/surface in the gauge?

yente/app.py
```diff
@@ -78,9 +104,13 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
         crontab=settings.CRONTAB,
         auto_reindex=settings.AUTO_REINDEX,
     )
-    settings.CRON = aiocron.crontab(settings.CRONTAB, func=cron_task)
-    yield
-    await close_provider()
+    settings.CRON = aiocron.crontab(settings.CRONTAB, func=refresh_catalog_cron_task)
+    metrics_task = asyncio.create_task(_run_every(60, update_metrics_task))
```
jbothma: why is this a custom periodic thingy instead of merely an aiocron task?

yente/search/indexer.py
```diff
@@ -213,6 +214,17 @@ async def index_entities(
...
+    # Always overwrite metadata: a fresh index has none, and a cloned index
+    # inherits the source's _meta (mappings can't be overridden in a clone request).
+    #
+    # Dates in ES documents would normally be stored as timestamps, but index
+    # metadata is just an opaque blob — ES doesn't parse or process it. For
+    # consistency we store values here exactly as we got them from the catalog.
+    index_metadata: Dict[str, Any] = {}
+    if dataset.model.last_export is not None:
+        index_metadata["last_export"] = dataset.model.last_export.isoformat()
+    await provider.set_index_metadata(next_index, index_metadata)
```
jbothma: this should only happen after aliasing, shouldn't it? Or am I confused?

aliasing but somehow failing to update metadata is ok - we get alerted and can check logs. updating metadata but failing to index and alias means we're think we're fine (when cron runs) but we're not fine.

jbothma (commented):
Added a couple of comments. 

generally, I think this approach looks great. for alerting on a stale index, it seems we can use `condition_monitoring_query_language` to configure an alert policy and alert if it's older than, say, 10h. No need to care whether the current index matches catalog version - we just want to know that we're `release period + max index time + margin` old.

nice nice

yente/search/indexer.py
leonhandreke: the metadata is something attached to the index, which contains one version of the data. The alias points to one of these versions, but has no metadata attached to it. The metric update function will look at which index is currently pointed to by the alias.

I pulled the metadata writing down *after* the indexing, because only then is it true that the index contains the data at the stated version. The call to `update_dataset_version_metric` is after the alias rollover happened, so that it picks up on the new index.

yente/app.py
leonhandreke: I had this be an aiocron, but that has a max frequency of 1 minute and is also another abstraction on top of asyncio. Currently, we run every 60s, but we might want to do more often in the future? No strong opinions here. WDYT?

yente/search/indexer.py
jbothma: yeah sorry I was using aliasing as proxy for "we've successfully indexed". Thanks for the move.

yente/app.py
jbothma: it feels like out of interest, why are we even setting it periodically and not just when it changes?

But if it actually needs to be set periodically and not just on change, it feels like just using the cron lib is simpler until we actually need more frequent setting.

As a very far aside, the sort of database designed for a running app holding statistics like this, called mnesia built into the Erlang runtime, is what klarna was originally built on. It was not designed for... a payment system's application data. And that resulted in a whole series of very interesting talks about making the wrong tool scale because migrating a running payment system away from it felt too hard :D

jbothma (approved):

yente/app.py
leonhandreke: The update code in `indexer.py` only gets called on reindex, i.e. not on startup and never for instances that don't have auto-reindex configured. I couldn't think of a better way - or do you think this could/should be something that the reindex job does maybe?

jbothma: oh that's really fun. yeah I can't think how to solve it so that each instance sets the gauge.

is it necessary for each instance to set the gauge or only whatever it is that updates the index? perhaps it's additional faff to set up otel for the cron job so having the api servers publish it is nice.

so yeah I think publishing every minute is fair. If we want it tighter so that the gauge literally means "this is the date stamp of the data we're serving in this response", the query itself should fetch the metadata on each request. Then there could be a little check if we've published this timestamp to otel yet.

So far I've only been thinking about this gauge as "roughly how stale is the index?"

leonhandreke: Yeah, same. I think for actually figuring out "what index was active at the time we served this specific request", we have the audit log.

PR #1136, opened 2026-05-08
search: index compound name tokens to match names with omitted spaces

leonhandreke:
## Summary

A query like "vladimirputin" (no space) arrives as a single token to the Elasticsearch analyzer. Fuzziness only applies within individual tokens, so it can't bridge the gap to the two indexed tokens "vladimir" + "putin" — the edit distance is just too large.

The fix: add an index-time shingle filter (`name-token-concat`) with an empty token separator to `osa-names-index-analyzer`. So "Vladimir Putin" gets indexed as `["vladimir", "putin", "vladimirputin"]`, and a query for "vladimirputin" finds the compound token directly.

We use a separate `search_analyzer` (plain `osa-analyzer`) so that query-time analysis is unaffected — otherwise a normal "Vladimir Putin" query would also generate a "vladimirputin" token that must be present, breaking matches against names like "Vladimir Vladimirovich Putin".

`max_shingle_size` is set to 2 (two-part concatenations). Raising to 3 would also cover three-part names at the cost of a larger index.

Requires a re-index to take effect.

## Known weakness

This only bridges *adjacent* indexed tokens. If a query both omits spaces *and* skips a middle name part — e.g. "vladimirputin" against an indexed "Vladimir Vladimirovich Putin" — we still can't match, because no shingle of size 2 over the indexed tokens produces "vladimirputin". The test `test_match_name_without_spaces` had to use "vladimirovichputin" instead (two adjacent tokens in the eu_fsf dataset entry) to exercise the fix.

Raising `max_shingle_size` doesn't fully solve this either — it would help for full omissions across N adjacent parts, but not arbitrary skips.

### DIFF
--- a/tests/test_match.py
+++ b/tests/test_match.py
@@ -202,6 +202,25 @@ def test_id_pass_through():
     assert res["query"]["id"] == "rotenberg"
 
 
+def test_match_name_without_spaces():
+    # A name query with spaces omitted ("vladimirovichputin") is a single token to the query
+    # analyzer, so ES fuzziness can't bridge the gap to the separate indexed tokens. We
+    # solve this at index time by also indexing compound tokens for adjacent name parts,
+    # and this test verifies that.
+    # The test also shows a limitation of our approach: the eu_fsf dataset only contains
+    # the full name "Vladimir Vladimirovich Putin", so we can't match "vladimirputin"
+    query = {
+        "queries": {
+            "a": {"schema": "Person", "properties": {"name": ["vladimirovichputin"]}}
+        }
+    }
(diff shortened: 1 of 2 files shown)


pudo:
I like this solution, but it does mean that now the expensive fuzzy matching is running on an extra name. Don't know how evil that is. The possible but less good alternative would be to make a keyword field like `name_variants` where we generate this and a few other transpositions (eg.: sorted name tokens) and ten query that using the same transposer. That would do less fuzzy but perhaps allow us to cover off a few more failure modes.

leonhandreke:
Superseded by https://github.com/opensanctions/yente/pull/1138

PR #1138, opened 2026-05-11
Index concatenated names as a keyword field

leonhandreke:
When a user queries a name with the spaces omitted (eg. "vladimirputin"), ES sees a single token and the per-token fuzziness can't bridge the gap to the separate indexed tokens ("vladimir", "putin"). So we miss matches that a human would consider obvious.

Follow-up to #1136 based on @pudo's suggestion: instead of running the expensive fuzzy matching on an extra synthetic concatenated name, generate a keyword `name_variants` field at index time and query it exactly with the same transposer.

Right now the only variant is "concatenated name parts" (so "vladimir putin" -> "vladimirputin"), but the structure leaves room for other transpositions later (eg. sorted tokens) without paying fuzzy cost.

### DIFF
--- a/yente/search/queries.py
+++ b/yente/search/queries.py
@@ -14,7 +14,8 @@ from yente import settings
 from yente.logs import get_logger
 from yente.data.dataset import Dataset
 from yente.data.util import entity_weak_names, index_symbols
-from yente.search.mapping import NAME_SYMBOLS_FIELD, NAMES_FIELD
+from yente.search.indexer import build_name_variants
+from yente.search.mapping import NAME_SYMBOLS_FIELD, NAME_VARIANTS_FIELD, NAMES_FIELD
 from yente.search.mapping import NAME_PART_FIELD, NAME_PHONETIC_FIELD
 
 log = get_logger(__name__)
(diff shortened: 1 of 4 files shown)


yente/search/queries.py
```diff
@@ -153,6 +154,12 @@ def names_query(entity: EntityProxy) -> List[Clause]:
 
     seen: Set[str] = set()
     for name in Name.consolidate_names(name_objs):
+        # Name variants are keyword renditions of common name variations that people expect to
+        # match but that the ES fuzzyness does not catch.
+        # Example: "vladimirputin" (without a space) does not get matched by the per-token fuzzyness
+        name_variants = build_name_variants(name)
+        shoulds.append(tqs(NAME_VARIANTS_FIELD, name_variants, boost=2))
```
leonhandreke: This is the line that I'd like a review on. Especially `tq` vs `tqs` (AND vs OR I think?) and the boost factor.

leonhandreke:
Could one of you take a quick look at this? I think I'd like a second pair of eyes on the line highlighted above before merging & releasing.

yente/search/queries.py
pudo: This seems correct: it's an OR on various names. Maybe one option might be to collect the full set for all the names and make that set into the filter?

PR #1140, opened 2026-05-12
search: fuzzy name match via SymSpell deletion neighborhood

leonhandreke:
## Summary

Profiling a `Vladimir Putin`-style `/match` query against the full production index showed that ~28 ms of the ~54 ms `took` was spent in `search.rewrite_time` — Lucene rewriting the `fuzziness: AUTO` clause on the `names` match by intersecting a Levenshtein automaton with the term dictionary FST. The cost scales with how wide the `names` term dictionary is (currently ~1.2M unique tokens in `name_parts`), which is exactly the direction we're growing.

This PR moves that cost from query time to index time:

- A new `name_parts_fuzzy` keyword field stores, for each normalized name token, the token itself plus every string obtained by deleting one character (SymSpell deletion neighborhood at depth δ=1).
- At query time, `names_query` generates the same expansion for each query token and looks it up with a plain `terms` query as an additional `dis_max` variant alongside the existing exact/phonetic/symbol channels.
- The `fuzziness: AUTO` clause on the `names` match is dropped — per-token fuzzy recall now lives in `name_parts_fuzzy`.

This makes fuzzy matching an inverted-index intersection instead of a per-shard term-dictionary scan, and the recall guarantee at edit-distance 1 (transpositions included) is provable per token rather than capped by Lucene's `max_expansions` top-N heuristic.

I haven't benchmarked this against the n-gram alternative below — the choice here is based on profile-driven reasoning about where the cost lives, not measured A/B. If we want to make a defensible choice we should index both into a side cluster and compare query cost and recall on real workloads.

## Alternatives

The Elastic-recommended way to do fuzzy name matching natively is **character n-grams**, e.g. trigrams. The idea: at index time, split each name into overlapping 3-character chunks — `Vladimir` becomes `vla`, `lad`, `adi`, `dim`, `imi`, `mir`. Store every chunk as a search key in the inverted index alongside the doc's other fields. At query time, do the same to the query string and find docs whose chunks overlap enough. Two strings that differ by a typo or two still share most of their chunks, so this gives you a soft "approximately the same name" recall channel. Elastic supports this directly via the `ngram` token filter and the standard analyzer chain.

This is a well-trodden recipe with real strengths: recall degrades gracefully with edit distance (a name with 3 typos still shares plenty of trigrams), and it naturally handles substring matches like `Vlad` ↔ `Vladimir` (both contain `vla` and `lad`). The downsides are practical:

**Big posting lists.** Elastic's inverted index works by, for each search key, storing the list of all doc IDs that contain it — that list is called a "posting list". When a query fires, Elastic walks the posting lists for the query's keys and finds the docs that show up enough times across them. Trigrams like `vla`, `dim`, `mir` appear in a huge fraction of names in the index, so their posting lists are tens or hundreds of thousands of doc IDs long. To answer a query for `Vladimir`, Elastic has to walk all six of those long lists and merge them. That's exactly the kind of `next_doc` work that already shows up as the dominant cost in our query profile (~9.75 ms on the main query). Trigrams would make that worse, not better.

**Threshold tuning.** You need to set `minimum_should_match` to decide how much overlap counts as "approximately the same name". Too low and the scorer drowns in candidates; too high and real matches drop off. There isn't one good value for all name lengths.

The deletion-neighborhood approach in this PR has the opposite shape: each key (like `vladimi` — `Vladimir` with one character missing) is very specific, so its posting list is tiny — only the few names that happen to delete to exactly that string. Walking it is cheap. The cost is that we get a hard cutoff at edit distance δ instead of graceful overlap, and we need a custom indexer step instead of letting Elastic's analyzer do it.

## Recall scope

Depth δ=1 covers all edit-distance-1 typos (sub/ins/del) and single transpositions, per token. Edit-distance 2 is **not** covered — there's a test case (`Viadimit Pufim`) documenting this gap. Extending to δ=2 widens recall at the cost of more keys per token.

### DIFF
--- a/tests/test_match.py
+++ b/tests/test_match.py
@@ -204,30 +204,93 @@ def test_id_pass_through():
 
 def test_fuzzy_names():
     """Test that fuzzy retrieval from the index works."""
-    query = {
-        "queries": {"a": {"schema": "Person", "properties": {"name": "Viadimit Putln"}}}
-    }
 
+    # With MATCH_FUZZY off, even a single-character typo on any name token should
+    # prevent the entity from being retrieved.
     with mock.patch("yente.settings.MATCH_FUZZY", False):
-        # We need to set a lower threshold to get logic-v2 to score it high enough.
-        # That's okay, we care about testing the fuzzy retrieval from the index,
-        # not the details of the scoring algorithm.
-        resp = client.post(
-            "/match/default", json=query, params={"algorithm": "best", "threshold": 0.2}
-        )
-        data = resp.json()
-        res = data["responses"]["a"]
-        assert len(res["results"]) == 0, res
-
-    with mock.patch("yente.settings.MATCH_FUZZY", True):
-        # The result scores quite low, so we need to set a lower threshold to get a result
+        # Edit-1 typo on each token. Both tokens have to be typo'd, otherwise an
+        # exact match on the untouched token alone is enough to retrieve Putin
+        # even without fuzzy matching. The typos are also chosen so their
+        # metaphones (FLTMT, PTNK) differ from those of the original ("Vladimir",
+        # "Putin" → FLTMR, PTN), so the phonetic channel doesn't retrieve Putin
+        # either.
         resp = client.post(
             "/match/default",
-            params={"threshold": 0.2, "algorithm": "logic-v2"},
-            json=query,
+            json={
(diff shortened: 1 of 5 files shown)


leonhandreke:
We'll likely be going with n-grams

PR #1147, opened 2026-05-22
Add PEP 751 lockfile and Dockerfile lock stage

pudo:
## Summary

First step of the SBOM / EU Cyber Resilience Act workstream. Full design in [`plans/sbom-and-cra-artifacts.md`](./plans/sbom-and-cra-artifacts.md) (also part of this PR).

- **`pylock.toml`** — PEP 751 lockfile, pinned runtime deps for `linux/amd64`, 132 packages with hashes. Generated via `pip lock` inside the same `python:3.12-slim` interpreter we ship.
- **`Dockerfile`** — new `lock` stage sibling to `build`, sharing the `python_image` ARG. Not part of the runtime image. Build / runtime stages unchanged.
- **`Makefile`** — `make lock` builds the `lock` stage and regenerates the file. No more ad-hoc env recreation.

## Not yet in this PR

This commit produces the lockfile but does **not** consume it. Follow-up commits on the same branch will:

1. Lock the `dev` / `docs` dependency groups (`pylock.dev.toml`) — pip's `pip install -r pylock.toml` doesn't support extras / dep groups yet, so per the PEP convention these get separate files.
2. Decide on `linux/arm64` strategy — PEP 751 lockfiles are single-platform; the current file is amd64-only.
3. Switch `Dockerfile` and `.github/workflows/build.yml` from `pip install -e .` to installing from the lockfile(s), with pip pinned to `>= 26.1`.

Steps 2 and 3 onward (SBOM generation, signing, SLSA provenance, GitHub Release automation, VEX) are scoped in the plan doc.

## Test plan

- [ ] `make lock` produces a 132-package `pylock.toml` identical to the committed file
- [ ] Existing `docker build .` and CI still pass (Dockerfile change is purely additive — new stage is not in the default build target chain)
- [ ] No regression in image size or runtime behaviour

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,19 @@
-# Define the base image to use for both stages
+# Define the base image to use for all stages
 ARG python_image=python:3.12-slim
+
+# ------------------------------------------------------------------------
+# Stage: Lock — environment used by `make lock` to produce pylock.toml.
+# Not part of the runtime image. Shares the python_image ARG so the
+# lockfile is resolved against the same interpreter we ship.
+# ------------------------------------------------------------------------
+FROM ${python_image} AS lock
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    build-essential \
+    pkg-config \
+    libicu-dev \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+RUN pip install --no-cache-dir 'pip>=26.1'
 
 # ------------------------------------------------------------------------
 # Stage 1: Build and install dependencies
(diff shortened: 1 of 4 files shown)


pudo:
Deprioritizing this in favour of #1148, which approaches the SBOM/CRA goal without requiring a lockfile.

Reasoning: scanning the built image with `syft` produces an SBOM that describes exactly what shipped, regardless of how resolution happened — so a lockfile isn't a prerequisite for either SBOM accuracy or CRA conformance. The lockfile is still genuinely useful for build reproducibility, forensic rebuilds, and security-review caching, but those are independent goals.

Leaving this PR open as a parking lot. Findings from the exploration (multi-arch lockfile cost, dep-group splitting, scanner-tooling gaps for PEP 751) are captured in §8 of the new plan doc so we don't have to re-discover them if/when build reproducibility becomes a priority on its own merits.

PR #1149, opened 2026-05-22
match: retrieve & score at most MAX_MATCH_CANDIDATES (default: 500)

leonhandreke:
This prevents slow responses and OOMs for high values of `limit`.

Open question: do we want to set the default to very high and only modify this in our prod deployment?


pudo (approved):
It eats at my pride a bit but I don't think it'll do the same thing to recall. Can you run the report once to check?

leonhandreke:
This only kicks in for values of `limit` >= 50. So I think running the report with the default of `limit = 5` shouldn't even trigger this?

pudo:
Oh, that's totally true. Let's ship it.

PR #1158, opened 2026-05-28
test: per-test fixture isolation with session-scoped index loading

leonhandreke:
Right now the test suite loads the full eu_fsf catalog once at session start and every test runs against it — which means tests aren't isolated and the suite requires a live internet connection to fetch from `data.opensanctions.org`.

This PR is a first step towards independence from the internet. The approach:

- Each dataset fixture (eu_fsf, parteispenden, zala) has a session-scoped loader that indexes its data into a fixture-specific ES index alias once per session. The function-scoped fixture then just monkeypatches `ENTITY_INDEX` to point at that alias for the duration of the test. This avoids re-indexing on every test function and prevents fixtures from stepping on each other's toes.
- Tests now declare which dataset they need via `@pytest.mark.usefixtures(...)` rather than all sharing a single global catalog.
- Added a self-contained `zala` fixture dataset (91 entities, 14 targets — Alexander Zakharov + LLC CST / Zala Aero sanctions cluster) as a local alternative to the live OpenSanctions data for tests that just need something to match against. There's a `generate.py` next to the fixture that documents how it was built and can regenerate it from any live yente instance.
- Manifest YAML files (`manifest.yml`, `sanctions.yml`) are replaced by inline `Manifest` constants in the test code, so there are no external files to keep in sync. `patch_yente_catalog(manifest)` async context manager centralises the save/restore of `yente.data._catalog`.

**Not fully migrated to be internet-independent yet** — the `live_catalog_eu_fsf` fixture and a handful of tests still hit `data.opensanctions.org` to fetch the real eu_fsf dataset. Posting this to validate the direction before going through each of those one by one.


leonhandreke:
@jbothma  could I get your thoughts on the general approach? The idea here is to load all test datasets into the index under different prefixes (session scope) and then only swap out the prefix for each test (function scope). `conftest.py` is probably the interesting place to look at. WDYT?

Also, thoughts on checking in the `generate.py` script even though we're not going to be using it often?

tests/conftest.py
```diff
@@ -17,51 +16,196 @@
...
 from yente.search.indexer import update_index
 from yente.provider import with_provider, close_provider
+from yente.data.manifest import Catalog, Manifest
+import yente.data
+
+
+@asynccontextmanager
+async def patch_yente_catalog(manifest: Manifest):
+    old_catalog = yente.data._catalog
+    yente.data._catalog = await Catalog.load(manifest)
```
jbothma, line 33: is this a singleton we set somewhere? I can't find it in the existing code or how this would be accessed.

if it is, it does feel iffy. Would monkeypatching get_catalog() be less surprising?

jbothma (commented):
I added one question - I might just be missing something obvious.

Generally I _love_ the live_catalog_ naming, I _love_ checking in generate.py to source countrol, and I think it looks like a very fair approach to isolating data for tests and a good use of what pytest has to offer in this respect - I learned a few things reading this.

Maybe a bit of a "fixtures strategy" comment at the top of conftest so it's easy to discover could make sense, just to be like "if you're adding a test, decide which data you need, and select it using x. if you need additional fixtures, add them by doing xyz"

tests/conftest.py
leonhandreke: Yes, iffy as hell. Before it lived on the `Catalog` class, and despite that not using the `global` keyword, I think it was just as bad, if not worse.

The problem with monkeypatching get_catalog is the way it in which monkeypatching works: you patch where it's imported, and it's imported and used in a bunch of different places. So the best thing I could think of was to set the variable directly, but keep the patching in one place.

leonhandreke: Ah, and to answer your question: It lives in `yente/data/__init__.py`, where `get_catalog` sets it.

leonhandreke:
Thanks for your thoughts! I added a little comment to the top of conftest, as you suggested.

leonhandreke:
@jbothma adding you again just to get some quick feedback on the singleton comment you made. I want to make sure I'm not making it worse than before :)

tests/conftest.py
jbothma: lol I missed you introducing _catalog because I didn't even ready to the bottom. Just assume all the changes were tests. Shocking.

The only better way I can come up with is to treat _catalog more explicitly like a cache and add a Catalog.reset() or Catalog.clear() or something, then the fixture could reset the catalog, patch settings.MANIFEST, and the next get_catalog() would be cool, or something. 

Is that a lot of scaffolding just to make sure patch_yente_catalog() stays in sync with get_catalog()? Maybe pydantic and mypy give enough safety that this is fine

jbothma (approved):

jbothma:
I replied with an idea so patch_yente_catalog is aesthetically less iffy, but practically I think this is fine.

tests/conftest.py
leonhandreke: The thing is that I don't want to write an actual manifest file, so I don't want to override `settings.MANIFEST`. I want to set the `Catalog` structure directly.

I guess I'll keep it like that for now. The `global _catalog` makes it clear that this is very internal, and a global singleton (which I like more than something attached to a class, somehow).

PR #1174, opened 2026-06-09
Check checksum

jbothma:
Fixes https://github.com/opensanctions/operations/issues/2543

Just double checking before making non-draft

- formats catalog fixture in one commit, before fixing data sha1 in subsequent commit
- checks checksum if available in a resource
  - doesn't add a manifest option to specify checksum for local files or entities url hardcoded in the manifest datasets section
- for simplicity, the checksum is always generated, and always checked after iterating over all the data, but before aliasing
  - it's only possible to check after iterating over the http stream, which is the default configuration
- warns if using resource as source but checksum isn't available (when enabled)
- includes option to disable checksum check - the use case I have in mind is someone serving from an internal server, perhaps custom data, and doesn't have their checksumming sorted, and doesn't want to be forced to fix that to upgrade yente. I think we should introduce a changelog item if we want to make it obligatory.

@leonhandreke I don't think it's a mad rush to get this in - totally ok to hold off until the upcoming release is out. I think we'll want to run this for at least a couple of days before releasing anyway.

- [x] is this noticably slower than `main`?
  - yes, 10% slower for sample of 1


jbothma:
I happened to run `yente reindex` at `2026-06-10T08:46:12.457904Z`

at `Jun 10, 2026, 8:48:25 AM UTC` we uploaded `data.opensanctions.org/artifacts/default/20260610065433-mhq/index.json`

at `Jun 10, 2026, 8:45:21 AM UTC` we uploaded `data.opensanctions.org/datasets/20260610/default/entities.ftm.json`

We calculated the hash for the new file correctly, but expected the old hash because the index job started before the fresh catalog was published.

So I was really unlucky - also I think it only happened because there was no index before. If there was an index, the stale catalog would mean we don't bother updating.

In production, until we implement https://github.com/opensanctions/operations/issues/2587, this might happen, but the next run will likely succeed. after 2587.

I think this is rare and recoverable enough that it doesn't have to wait for 2587 but we should probably do that next.

jbothma:
benchmark results

command: `yente checksum`

main branch

```
real	14m37.563s
user	5m8.794s
sys	0m9.339s
```

feature branch

```
real	15m57.893s
user	6m59.932s
sys	0m8.677s
```

10% slowdown

yente/data/loader.py
```diff
@@ -77,44 +89,100 @@ async def read_path_lines(path: Path) -> AsyncGenerator[Any, None]:
...
+            yield chunk
+
+    def hexdigest(self) -> str:
+        return self.digest.hexdigest()
+
+
+async def split_json_lines(chunks: AsyncIterable[bytes]) -> AsyncGenerator[Any, None]:
+    """Split byte chunks on newlines and yield each non-empty line as parsed JSON."""
+    buf = b""
+    async for chunk in chunks:
```
leonhandreke, line 109: We could use `LineDecoder` from `httpx` here and drop this custom splitting + the tests for it? It's private but meh.

yente/data/dataset.py
```diff
@@ -49,7 +50,20 @@ def __init__(self, data: Dict[str, Any]):
         if self.model.load is None:
             self.model.load = not self.is_collection
         if self.model.entities_url is None:
-            self.model.entities_url = self._get_entities_url(data)
+            if self.model.path is not None:
+                self.model.entities_url = self.model.path.resolve().as_uri()
+            else:
```
leonhandreke, line 55: Would it make sense, just for consistency, to allow setting both `path` and `entities_checksum`? I think if `path` is set, the else-path never happens.

leonhandreke (commented):
Nice!

yente/data/dataset.py
jbothma: What's the use case? data loaded over nfs? I like having a warning so that anyone generating their own catalog gets nudged to fill in the checksum field without it breaking hard. But I don't think someone reading a static file should get that warning. That, and not really seeing the use case, made me go with this implementation.

Not that I love 7 levels of indentation.

yente/data/loader.py
jbothma: this? https://github.com/encode/httpx/blob/master/httpx/_decoders.py#L321

I'm not sure I see how it iterates - it seems to work on a string?

yente/data/dataset.py
jbothma: so I think my main hesitation is that unless we drop the warning, I don't think it simplifies the logic that much. While it expands the config and perhaps support surface. But shout if I'm not seeing how it'd be better - I might be fixed on one track

leonhandreke: Those are fair point. I was mostly thinking less indentation, less control flow. But I fully agree on the warning. Let's keep it like this!

yente/data/loader.py
leonhandreke: I explored this a bit and came to the conclusion that it's not really less code - this seems pretty safe in any case, even though I'm not super happy with having to implement this ourselves.

PR #1219, opened 2026-07-24
Serve an ETag on GET /catalog to enable conditional requests

### LINKED ISSUE #1202: Serve an ETag on GET /catalog to enable conditional requests
## Motivation

Long-lived API consumers (e.g. the `yente-client` MCP server) poll `GET /catalog` to keep dataset metadata fresh — in that context, dataset **titles** are re-resolved on a TTL to label screening results. Today the endpoint serves no cache validators:

```
$ curl -s -D - -o /dev/null https://api.opensanctions.org/catalog | grep -i "etag\|last-modified\|cache-control"
(nothing)
```

so every refresh is a full refetch of a response that changes only when a dataset version bumps (every few hours at most). By contrast, `data.opensanctions.org/meta/programs.json` already serves `ETag` + `Last-Modified` + `Cache-Control`, and clients can revalidate it with a bodiless 304.

## Proposal

Have `GET /catalog` (`yente/routers/admin.py`, `catalog()`) emit an `ETag` and honour `If-None-Match` with a `304 Not Modified`. The response is fully determined by the loaded datasets' `version` / `index_version` pairs, so a cheap strong validator is a hash over `(name, version, index_version)` for all datasets — no need to serialize the body to compare.

Optionally add `Cache-Control: public, max-age=<a few minutes>` so intermediary caches get a hint too.

## Client side

`yente-client` is growing a generic per-URL ETag revalidation path for `programs.json`; it activates automatically for any endpoint that starts returning an `ETag`, so `/catalog` support lands on the client side with no further changes there.

DMZ22:
Closes #1202.

### What

`GET /catalog` (and the `/manifest` alias) now emit a strong `ETag` and honour `If-None-Match` with a bodiless `304 Not Modified`, so long-lived pollers (e.g. `yente-client`) can revalidate cheaply instead of refetching an unchanged body on every TTL tick.

### How

- The response is fully determined by the loaded datasets' `(name, version, index_version)` triples, so the validator is a SHA-1 over those (sorted, so it's order-independent) — computed after `sync_dataset_versions`, without serialising the body to compare. It changes exactly when the response would.
- `If-None-Match` is matched against the ETag with support for the `*` wildcard and a comma-separated list; on a match the handler returns an empty `304` that still carries the `ETag`.
- A short `Cache-Control: public, max-age=60` is added as the issue suggested, so intermediary caches get a hint while still revalidating well within the "every few hours" version-bump cadence.

Kept to `yente/routers/admin.py`; the ETag/If-None-Match logic is factored into two small pure helpers (`catalog_etag`, `if_none_match`).

### Tests

Added to `tests/test_base.py`:
- `test_catalog_etag` — drives the endpoint through `TestClient`, asserting: `200` carries an `ETag` + `Cache-Control`; a matching `If-None-Match` returns a bodiless `304` with the validator; a stale validator returns the full body; `*` matches. It stubs the search provider (the endpoint only asks it for aliased index names via `sync_dataset_versions`), so it needs no live Elasticsearch, following the existing `test_catalog` pattern.
- `test_catalog_etag_helpers` — unit-covers the hash (order-independence, sensitivity to version / index_version changes, strong-validator quoting) and the `If-None-Match` matching (list + `*` + miss).

Happy to adjust the `max-age`, drop the `Cache-Control`, or move the helpers if you'd prefer them elsewhere.

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -1,7 +1,11 @@
+from types import SimpleNamespace
+from unittest.mock import AsyncMock
+
 import pytest
 
 from .conftest import (
     FIXTURES_PATH,
+    app,
     build_index_alias_name_for_fixture,
     client,
     patch_catalog_response,
@@ -10,6 +14,8 @@ from .conftest import (
 
 from yente import settings
 from yente.data.manifest import Manifest
+from yente.provider import get_provider
+from yente.routers.admin import catalog_etag, if_none_match
 from yente.search.indexer import update_index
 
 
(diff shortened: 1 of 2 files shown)


leonhandreke:
Hi, thank you for this contribution - looks good! I have a few small nits which I will merge as followup commits.

Issue #1001, opened 2026-01-22
[docs] Should we start discouraging batching or at least not encourage it?


leonhandreke:
From https://github.com/opensanctions/yente/issues/957

> Currently, I'm thinking that the ability to batch multiple queries in one request has maybe turned out to be one of the less successful decisions in the yente API design. For a few reasons:
> 
> 1. it complicates the semantics, because now you have parameters that are per-request but your really want them per-query
> 2. we're running on Cloud Run and there the model is to run many small instances and scale up only if needed (which I understand is not what you're doing). So really, it would be better to have more requests of roughly equal weight come in instead of some that are really fat.
> 3. should be easy to do on the client, and that's less complexity on the server (but maybe I'm living too much in async unicorn land an in practice, people are working with much less nice language support on their clients?)

We could at least make the examples not contain batch queries. I'd prefer to actually discourage it somewhere in the docs. I feel like this is a change we can make without causing a commotion. One day, we'll have a v2 of the API, but that day is surely not now.

Somewhat related: https://github.com/opensanctions/yente/issues/927

cc @goose-life

PR #1249, opened 2026-09-10
Stop showing a batch as the /match request example

toruiwasa:
Related to #1001.

[`docs/deploy/scaling.md`](https://github.com/opensanctions/yente/blob/d68b2af170da4b446cfe0e5f17ff33a61848b055/docs/deploy/scaling.md#L37) says not to batch queries into one request, but the `/match` API reference still shows a two-entity batch as its example. This reduces the example to a single query and adds a note that more than one query is still accepted but not recommended. The API itself is unchanged.

`examples/match_api.py` is changed in a separate commit, in case you'd rather leave `examples/` alone.

### DIFF
--- a/examples/match_api.py
+++ b/examples/match_api.py
@@ -26,9 +26,9 @@ EXAMPLE_2 = {
     },
 }
 
-# We put both of these queries into a matching batch, giving each of them an
-# ID that we can recognize it by later:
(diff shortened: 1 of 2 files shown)


leonhandreke:
Thanks for spotting this!

PR #686, opened 2025-03-14
Paginated adjacencent entities API

jbothma:
- [x] see if we can keep these new endpoints out of the docs until we've run them in prof for a bit.

Implementation for #649

adds an endpoint `/entities/{entity_id}/adjacent` providing a page of results for each property whose values can be entities (both outbound references like addresses, and inbound references like ownership) - this is useful for the first page of results e.g. to show on an entity web page. Supports further pagination for completeness but that is an unlikely use case.

Adds an endpoint `/entities/{entity_id}/adjacent/{propertyName}` providing paginated access, e.g. for further pages of results for a given id.

## Quirks

- **pagination only at first level**: interstitial entities like Ownership nest the other side of the relationship. We don't paginate that side because it doesn't make sense without additional pagination parameters, e.g. offset=10 should give the next 10 items of the ownerships of the root person, but if the assets were paginated, how would you get the first 10 assets of those ownerships? We assume that each interstitial entity would link one or very close to one entity with another.
- **missing entities**: In the current nested endpoint, if an outbound related entity is missing, we'll still include its id in the  response alongside the nested entities that did exist. In the new adjacent endpoints, we drop the missing IDs because the page limit means we sometimes only fetch some of the referenced entities. Dropping the IDs means we can use elasticsearch pagination for outbound references just like inbound references, and don't have to track which IDs were actually requested and could be expected to be returned.

## Examples

***note*** that the all-props responses now nest limit and offset alongside total for each prop to make the types and code simpler - less marshalling

```json
time curl localhost:8000/entities/Q10891456/adjacent?limit=1 |jq .
{
  "entity": {
    "id": "Q10891456",
    "caption": "National People's Congress deputy",
    "schema": "Position",
    ...
  },
  "adjacent": {
    "occupancies": {
      "results": [
        {
          "id": "wd-00058383f70e1306566103de8495829d2ba24712",
          "caption": "Occupancy",
          "schema": "Occupancy",
          "properties": {
            "holder": [
              {
                "id": "Q16905522",
                "caption": "栗俊平",
                "schema": "Person",
                ...
              }
            ],
            ...
            "post": [
              "Q10891456"
            ]
          },
        ...
        }
      ],
      "total": {
        "value": 10000,
        "relation": "gte"
      }
    }
  },
  "limit": 1,
  "offset": 0
}

real	0m0.080s
```

```json
 time curl localhost:8000/entities/Q10891456/adjacent/occupancies?limit=1 |jq .
{
  "results": [
    {
[...]

PR #1250, opened 2026-09-11
Test deep nesting through a Family edge

### LINKED ISSUE #80: Re-instate deep nesting tests
After migrating OpenSanctions to use externals, the wd_curated dataset ended up being empty. That's what all tests for yente were written against. So I've now switched it over to eu_fsf, but eu_fsf doesn't have family ties. Need to find that somewhere to test deep nesting of entities again.

toruiwasa:
Closes #80.

Adds a test that walks the nested response across a `Family` edge: the root person, the four `Family` entities naming them as a relative, and the person on the other side of each. The `zala` fixture already has these ties, so no fixture changes.

The general case has been covered since #686. `test_entity_nested` now walks `Person → Payment → Organization` on `parteispenden`. What was missing is the `Family` case this issue names, and `zala` is the only fixture with it.

The `, res.text` on the status assertion is a separate commit. Most assertions under `tests/` attach a message but this file mostly doesn't, so drop it if you'd rather keep the file consistent.

### DIFF
--- a/tests/test_entity.py
+++ b/tests/test_entity.py
@@ -50,6 +50,7 @@ def test_entity_nested_family():
 
     families = props["familyRelative"]
     assert len(families) == 4
+    assert isinstance(families[0], dict)
     fam = by_id(families, "ofac-6d023267c742427bc92220ad97daca2e17b136ff")
     assert fam["schema"] == "Family"
     assert fam["properties"]["relationship"] == ["Family member of"]
--- a/tests/test_entity.py
+++ b/tests/test_entity.py
@@ -39,6 +39,35 @@ def test_entity_fetch():
     assert sanc["schema"] == "Sanction"
 
 
+@pytest.mark.usefixtures("zala_test_dataset")
+def test_entity_nested_family():
+    """Deep nesting across a Family edge: the root person, the Family entities
+    naming them as a relative, and the person on the other side of each."""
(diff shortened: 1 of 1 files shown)


leonhandreke:
As you wrote in the PR description, `test_entity_nested` already covers this case, so I don't think we need an additional one. We try to keep the test suite comprehensive but lean.

I don't quite know how you ended up wanting to tackle this issue, but if it's just an LLM looking for low-hanging fruit: We generally discourage low-effort contributions that are LLM-driven and contain very little to no human effort. If you're a human eager to join the community and contribute, that's something we absolutely welcome and I'm sure we can find some issues that would be fun and rewarding to work on!

toruiwasa:
Thanks for the honest feedback, and sorry for the noise. I wanted to start with something small to earn a bit of trust before taking on anything bigger, and this was the wrong pick for that.

To be upfront: I do use a coding agent, and I use it a lot, including for the English in this comment. The choice of issue and the decision to submit were mine. If that doesn't fit how you want contributions to happen here, I'll adjust.

I found yente because I worked in risk and compliance for about a year and was looking for an open source project close to that area. Reading through the open issues, I found most of them are still being thought through, or sit close to work you're doing right now, so I didn't want to get in the way of that. If there's anything you'd be happy to hand to an outside contributor, I'd like to take it. Bug reports, tests or docs are fine. It doesn't have to be a feature. And if you'd prefer I give you a heads-up on the issue before sending a PR next time, I'm happy to do that.

PR #1257, opened 2026-09-13
Retrieve /match candidates per comparable name part

### LINKED ISSUE #1252: Candidate retrieval penalises entities with many aliases via the names text clause
## Problem

The `/match` candidate query should treat a name token as a retrieval criterion: an entity with 120 aliases containing "Hezbollah" is not less "Hezbollah" than an entity with two. The heavily merged entities are the ones a screening API most needs inside the candidate window.

The keyword fields already behave this way (`name_parts`, `name_phonetic`, `name_symbols` are sets with norms off, so a term scores IDF-only). The `match names` clause in `names_query` does not: `names` is a text field, so BM25 term frequency and document length apply, and `weak_length_norm` (b=0.25) only softens the effect. It is also the highest-boosted clause in the query (×3).

## Evidence

`_explain` for the query `{"schema": "Organization", "properties": {"name": "Hezbollah"}}` on three entities that all carry the exact token (local ES, index 018):

| entity | aliases | `names:hezbollah^3` | `names.ngrams` | `name_parts` | `name_symbols` | `name_phonetic` | total |
|---|---|---|---|---|---|---|---|
| Ansar-e Hezbollah | 8 (dl 31) | 58.3 | 13.3 | 17.8 | 8.7 | 5.2 | 92.2 |
| Bahrains Hezbollah | 2 (dl 6) | 37.3 | 12.7 | 17.8 | 8.7 | 5.2 | 70.6 |
| Hizballah (NK-npdZSrDYbpdVnmUndsGzzA) | 122 (dl 856) | 25.4 | 13.0 | 17.8 | 8.7 | 5.2 | 59.0 |

The canonical Hizballah entity ranks 7th of the 7 exact matches. For the query "Ali", "Mohammad Ali Jafari" (74 aliases) ranks 53rd, outside the default 50-candidate window, behind dozens of single-alias entities that also carry the token.

In the tested single-alias queries no exact match is displaced by fuzzy noise, but where more entities carry all query tokens than fit the window ("Mohammed Ali": 201 exact matches, 50 slots), alias count and length decide which ones get scored. That is the wrong criterion.

## Proposal

Replace the `match names` (operator AND, ^3) clause with a `bool` `must` over `term name_parts` for each query part, same boost. It expresses the same thing ("document has all the query tokens"), on the field that is already symmetric with the query-side normalisation (see #1253), without tf or norms. No reindex needed.

Then re-check whether `upscore_large_entities` (used in `/search` only) is wanted for `/match`, or whether it becomes unnecessary.

The same tf/norm leak on the `names.ngrams` sub-field is handled by the n-gram rework in #1254 (`index_options: docs`, `norms: false` on the new sub-field).

## Gate

`contrib/validation_report` / `contrib/candidate_generation_benchmark` id_recall unchanged or better, measured by presence in the candidate window; additionally check the rank of the most-merged entity for the sparse organisation queries in `contrib/perf_bench/queries.json` (Hamas, Hezbollah, Wagner, Ali) before/after.

### LINKED ISSUE #1253: Two normalisation families in the match index: names/ngrams (ES analyzer) vs name_parts (rigour)
## Problem

Index-time and query-time normalisation must be symmetric for a term to be retrievable. The `/match` index currently has two normalisation families, each symmetric with itself but not with the other:

- **ES analyzer family**: `names` (text, `osa-analyzer`) and `names.ngrams` (`osa-ngram-analyzer`): standard tokenizer, lowercase, asciifolding, applied to raw name property values on both sides.
- **rigour family**: `name_parts`, `name_phonetic`, `name_symbols`: rigour tokenisation, latinisation, org-type rewriting, applied via `entity_names` on both sides.

`_analyze` vs rigour comparable parts for the same string:

| input | ES tokens | rigour parts |
|---|---|---|
| O'Neill | `o'neill` | `oneill` |
| Владимир Путин | `владимир`, `путин` | `putin`, `vladimir` |
| Абу Бакр аль-Багдади | `абу`, `бакр`, `аль`, `багдади` | `abu`, `bakr`, `al`, `bagdadi` |
| 株式会社トヨタ | `株`, `式`, `会`, `社`, `トヨタ` | `株式会社トヨタ` |
| Xi'an Aircraft Industry (Group) Co., Ltd. | `xi'an`, ..., `co`, `ltd` | `xian`, ..., `coltd` |

## Consequences

- **No cross-script fuzziness.** The n-gram clause on a Cyrillic query only matches Cyrillic aliases; a typo in a transliterated name is only recoverable via `name_phonetic`, which requires an ASCII form and ≥3 characters.
- Apostrophes, hyphen-joined and CJK names are tokenised two different ways; the `names` AND clause requires tokens rigour would have rewritten (`co`, `ltd` vs `coltd`).
- The per-alias n-gram clauses cannot be deduplicated across transliterations of one name.

## Fixes

The name-field half is covered by two issues; the direction in both is to stop querying raw-value ES-analyzed fields in `/match` and work from rigour output:

- **#1254** replaces `names.ngrams` with a trigram multi-field on `name_parts` and per-token clauses (cross-script, deduplicated across aliases).
- **#1252** replaces the `match names` clause with a conjunction over `name_parts`.

After both, `/match` would not touch `names` or `names.ngrams` at all; they stay for `/search`.

## Smaller asymmetries, same principle

Each can ship alone:

- **Identifiers.** `registry.identifier.clean` keeps spaces/dashes (`IMO 9181156`, `IMO9181156`, `9181156` are three terms) and the `osa-normalizer` only lowercases and asciifolds, so differently formatted identifiers never meet at retrieval. `rigour.ids` has per-format `normalize` functions; index a normalised form alongside the raw value and query both.
- **Dates.** `expand_dates` adds month/year prefixes at index time only, so a full-date query cannot reach a year-only record (the reverse works). Add the truncated forms of query dates as lower-boosted terms.
- **Consolidation.** The indexer calls `entity_names` with the default `consolidate=True`; the function's docstring says indexers should pass `False`. Parts sets are unaffected (a contained name adds no parts), symbol sets may be. Switch and add a test asserting the indexed symbols for a name and its contained short form.

### LINKED ISSUE #1254: Rework n-gram candidate retrieval: per-token trigrams on comparable parts
Non-vibe - this is what I asked the agent to do: 

 I want you to draft a document in plans/match-es-perf.md focussed on how the match API queries elasticsearch. The idea is to audit it against the following princples:

  - Recall is really important. This is a sanctions screening API, a false negative is a regulatory incident.
  - The ES call in match is a candidate generator, not the matching system. It only needs to generate candidates for in-code scoring, not precise relevance scores itself.
  - Symmetry between the normalisations applied at index and query time is really essential to this.
  - If an entity has many names, the retrieval criterion can be the name token. A person having 20 aliases all of which contain "John" doesn't mean they're more called John than a person with only one alias. This may particularly help with ngrams.
  - If we can pre-compute stuff during index to get query-time perf, that's really nice

---

Umbrella issue for the n-gram half of #1253 (normalisation families) and item 2 of #1252 (alias penalty). One reindex, one query change; the pieces have to be validated together because each changes what the others admit.

## Today

`names_query` emits, for up to five `representative_names`, a `match names.ngrams` clause on the whole alias string (`minimum_should_match: 70%`, boost 1.5). `names.ngrams` is a text sub-field of `names` built from raw property values with the ES standard tokenizer + trigram filter, default BM25 (tf, norms, positions on).

Problems:

- **Cost.** The n-gram clauses are 70–90% of ES time on many-name queries (local index 018, median `took`, ms: dense-putin 41 → 6 without them, long-name-string 62 → 7, names-ermakov-20 26 → 8, names-stroygaz-40 24 → 7). Five whole-alias clauses of 20–30 trigrams each, mostly the same trigrams repeated across transliterations of one name.
- **No cross-script fuzziness.** Trigrams are built from the raw script; a Cyrillic query only fuzzy-matches Cyrillic aliases. `name_parts` is latinised on both sides, the n-gram field is not.
- **Short tokens.** The `is_short` rule forces n-grams on single-word queries, where a ≤5-char word yields ≤3 trigrams. One trigram cannot express typo tolerance; it only widens the set ("Hamas": 99 → 10,992 total hits; the trigram `ali` alone is in 35,627 docs).
- **Alias count leaks into the score.** tf and length norm on the sub-field mean an entity's trigram score depends on how many aliases it has, not on whether a token matches.

## Proposal

**Mapping** (`yente/search/mapping.py`, reindex required):

- Add a `text` multi-field on `name_parts`: `name_parts.ngrams`, analyzer `osa-ngram-analyzer`, `index_options: docs`, `norms: false`. ES builds it from the comparable parts yente already writes, so there is no indexer change: one trigram bag per distinct latinised token, deduplicated across aliases, symmetric with the query-side comparable parts.
- Once `/match` no longer queries it, drop the `names.ngrams` sub-field (112 MB of postings; `names` itself stays for `/search`).

**Query** (`yente/search/queries.py`):

- Replace the per-alias `names.ngrams` clauses with one `match name_parts.ngrams` per unique comparable part of the representative names, parts of ≥5 characters only, `minimum_should_match: 70%` to start, boost to be tuned.
- Remove the `is_short` rule. Short tokens keep exact, phonetic and symbol clauses only, consistent with how the rest of the query treats them.
- `MATCH_FUZZY` keeps its meaning (n-gram clauses on/off).

## Prototype numbers

Per-part clauses over the representative names' comparable parts (≥5 chars) against the existing `names.ngrams` field, i.e. query change only, no reindex. Local ES 9.4.2, index 018, median `took` of 5 warm runs, ms, and overlap of the top-50 window with the current query:

| query | current | per-part | n-gram clauses current → per-part | top-50 overlap |
|---|---|---|---|---|
| names-ermakov-20 | 26 | 20 | 3 → 7 | 47 |
| names-stroygaz-40 | 26 | 11 | 5 → 3 | 11 |
| dense-putin | 31 | 12 | 5 → 5 | 36 |
| long-name-string | 105 | 38 | 1 → 19 | 4 |
| script-cyrillic-putin | 11 | 8 | 1 → 3 | 4 |
| pos-rotenberg | 12 | 9 | 2 → 4 | 48 |
| names-kim-12 | 13 | 11 | 5 → 1 | 40 |
| neg-john-smith | 3 | 3 | 1 → 1 | 34 |

With `track_total_hits: false` (#1251) on top: ermakov 14, stroygaz 10, putin 10, long-name 26.

The low overlap on stroygaz/long-name/cyrillic-putin is expected: those windows are currently filled by whole-alias trigram matches that the per-token form scores differently. Whether that is a recall change is exactly what the gate below has to answer; the exact-match check (entities carrying every query token) showed none displaced in either form.

## Open questions

- **Threshold per token.** Per-token clauses tolerate roughly one edit per token where the per-alias clause had an edit budget across the whole name. 70% of a 6-trigram token is 4, so one mid-token edit (kills 3 trigrams) fails. Options: 60–70%, or a length-dependent msm. Decide on the `positives_un_treated` fixture, which is the one that moved in #1139.
- **Missing spaces.** "vladimirputin" as one query token still matches trigrams of the indexed `vladimir` and `putin` parts (9 of 11 trigrams; trigrams never cross a token boundary on either side), so the case from `test_match_name_without_spaces` should keep working; verify.
- **Cutoff at 5 characters** vs 4: 4-char tokens give 2 trigrams; probably still noise.
- Boost relative to the `name_parts` term clause once the sub-field is IDF-only.

## Gate

- `contrib/validation_report` id_recall per fixture, measured by presence in the candidate window, before/after; no regression.
- `contrib/candidate_generation_benchmark` same.
- `contrib/perf_bench --output` on both trees for latency.
- Existing tests `test_fuzzy_names` and `test_match_name_without_spaces` must pass unchanged.

## Order

1. Query change against the existing `names.ngrams` (no reindex) to settle threshold and cutoff on the recall suites.
2. Mapping change, reindex, switch the clauses to `name_parts.ngrams`, re-run the gate (this is where cross-script fuzziness and the alias-independent scoring arrive).
3. Drop `names.ngrams`.

pudo:
Rebuilds how `/match` retrieves name candidates from Elasticsearch: one rigour normaliser on both sides, one unit of evidence per name part, and four explicitly named ways a query part may reach an indexed one. Closes #1252, closes #1253, closes #1254.

## Index

`build_indexable_entity_doc` now writes a third derived name field, `name_joined`: for every matchable name, the comparable forms of its parts concatenated without spaces (`alqaeda`, `vladimirputin`). It is a plain keyword set excluded from `_source`, like `name_parts` and `name_symbols`. The `name_phonetic` field and the `names.ngrams` trigram sub-field (with its `osa-ngram-filter`/`osa-ngram-analyzer`) are removed; the indexer no longer computes metaphones. The `names` text field and its `weak_length_norm` similarity stay for `/search`; `name_parts` keeps its normalizer because weak aliases share the field. Field sizes on the full `default` collection: `name_joined` 89 MB added, `name_phonetic` 28 MB and `names.ngrams` 117 MB removed; total store size 2,645 MB vs 2,657 MB.

## Query

`names_query` builds one `dis_max` (tie_breaker 0) per unique comparable name part across all of the query's names, taken from the same cached `entity_names(entity, is_query=True)` the scorer uses, so twenty transliterations of one name yield about as many clauses as one name has parts. Inside each `dis_max` are up to three channels; a document scores each query part once, whichever channel matched and however many aliases carry it, and the outer `should` sums over parts, so a document ranks by how many query parts it covers, weighted by part rarity. One `terms` clause on `name_joined` covers all space-less forms of the query names; weak aliases are queried exactly as before.

The four channels, in plain words:

- **exact**: the indexed part is the same comparable string (casefolded, latinised where the script allows it, diacritics and punctuation stripped). IDF-scored `term`.
- **fuzzy**: the indexed part is within Damerau-Levenshtein `AUTO` of the query part (0 edits up to 2 chars, 1 edit at 3–5, 2 edits from 6) with the first letter fixed; only when `MATCH_FUZZY` is on and the part has at least three characters. Run as a `constant_score` filter so a record with ten spellings of one part scores the hit once, worth a fixed `FUZZY_BOOST = 6.0` (about the exact score of a part shared by 25,000 records, so an exact hit on any but the most common tokens outranks an approximate one).
- **symbol**: rigour tagged both parts with the same known-name identity, nickname, org class etc. IDF-scored `dis_max` of `term`s on `name_symbols`, boost 0.9. The only bridge for scripts that are not latinised (Arabic, Han) and for variants beyond two edits (alexander/aleksandr).
- **joined**: a query name with its spaces removed equals an indexed name with its spaces removed. `terms` on `name_joined`.

Caps: `MAX_PARTS = 100`, `MAX_SYMBOLS_PER_PART = 30` bound the query at roughly 3,200 clauses (observed maxima: 22 parts, 29 symbols on one part). Primary names are visited before aliases so a truncated query keeps its primary name whole.

Two calibrations of clauses outside the name channels were needed once the ×3 `match names` clause was gone: `TYPE_BOOSTS[date]` goes from 3.0 to 1.0, because a year-only birth date at 3.0 scored more than an exact two-part name match and over a thousand records share any given year, so they filled the candidate window on their own (US Congress fixture: 98.6% → 86.2% before this change, 98.7% after). The fuzzy clause is wrapped in `constant_score` because a bare fuzzy clause is rewritten into an OR of its expansions and sums the ones a document carries, which handed heavily aliased records the very alias bonus this query removes ("Andrea Salinas" lost to a record with ten transliterations of Andrey scoring 40 on "andrea" alone). Using the `constant_score` *rewrite* instead would have been 3–5× slower on many-part queries because it enumerates every term in the edit budget rather than the top 200.

## Recall@50, before/after

Presence of the expected id(s) in a 50-candidate window; router-identical query construction (`entity_query`, filter_op OR, `DEFAULT_SORTS`, `track_total_hits: false`). "Before" is `origin/main` against the 018 index of `default`, "after" is this branch against a 019 index of the same collection built with this code (a slightly newer data release: 4,049,275 vs 4,049,113 documents). Fixtures from `contrib/validation_report/build/fixtures`.

| fixture | n | recall@50 before | after | median rank before | after |
|---|---|---|---|---|---|
| positives_merged_aliases | 200 | 85.0% (170) | 99.5% (199) | 2 | 1 |
| positives_name_bench | 246 | 95.9% (236) | 97.6% (240) | 1 | 1 |
| positives_un_treated | 732 | 96.6% (707) | 96.6% (707) | 1 | 1 |
| positives_un_untreated | 732 | 96.6% (707) | 96.6% (707) | 1 | 1 |
| positives_us_congress_untreated | 1321 | 98.6% (1302) | 98.7% (1304) | 1 | 1 |

Treated UN fixture by whether the name was mangled: name changed 238, 96.2% → 96.2%; name unchanged 494, 96.8% → 96.8%. The 25 UN misses are shared by both queries (entities outside the window for both).

Misses that changed:

- merged aliases: 29 gained, 0 lost. Charles III, Elizabeth II, Dmitry Medvedev, Pope Francis, the Federal Penitentiary Service, Donald Trump ("Don"), Abu Nidal Organisation ("ANO") and other heavily merged records now rank first or near first for their own most-shared alias; previously the window was filled by single-alias records via BM25 on the `names` text field.
- name bench: 4 gained (Emanuel Macron, Emmanuel Macrone, Emmanuel Marcon → rank 1/1/3; Mark Cooney → Mark Carney rank 26), 5 lost: `rimaldiV nituP` and `Vladimir nitPu` (reversed strings, only trigrams could match these), `Kim Yong-chol` → Kim Yong-chun twice (two edits on a four-letter syllable, outside `AUTO`), `عبد الفتاح السري` → السيسي (Arabic single-part edit, first letter fixed and no NAME symbol).
- US Congress: 2 gained (Randy (Duke) Cunningham, Harry E. Mitchell, both rank 1), 0 lost.

## Latency

`contrib/perf_bench/perf_bench.py run --output …` from both trees, same index builds as above, 100 queries, medians of 3 warm runs.

| stage | p50 before | after | p95 before | after | max before | after |
|---|---|---|---|---|---|---|
| search (ms) | 8.4 | 8.4 | 41.9 | 35.0 | 99.5 | 71.7 |
| ES took (ms) | 5.0 | 5.5 | 38.1 | 31.1 | 93.0 | 67.0 |
| score (ms) | 7.5 | 6.5 | 70.7 | 48.2 | 423.0 | 543.7 |

| query | tags | took before | after | hits before | after |
|---|---|---|---|---|---|
| sparse-ali | sparse | 7 | 4 | 50 | 50 |
| sparse-hamas | sparse | 3 | 2 | 50 | 36 |
| sparse-hezbollah | sparse | 3 | 2 | 50 | 12 |
| sparse-wagner | sparse | 5 | 3 | 50 | 50 |
| sparse-putin | sparse | 7 | 5 | 50 | 50 |
| sparse-ivan-petrov | sparse | 6 | 5 | 50 | 50 |
| sparse-rosneft | sparse | 3 | 3 | 50 | 45 |
| sparse-gazprom | sparse | 5 | 3 | 50 | 50 |
| sparse-legalentity-bank | sparse | 4 | 2 | 50 | 50 |
| names-zakharov-7 | many-names | 34 | 20 | 50 | 50 |
| names-ermakov-20 | many-names | 28 | 58 | 50 | 50 |
| names-putin-10 | many-names | 34 | 31 | 50 | 50 |
| names-rotenberg-5 | many-names | 38 | 22 | 50 | 50 |
| names-deripaska-6 | many-names | 17 | 15 | 50 | 50 |
| names-kim-12 | many-names | 46 | 25 | 50 | 50 |
| names-rosneft-15 | many-names | 37 | 38 | 50 | 50 |
| names-stroygaz-40 | many-names | 40 | 40 | 50 | 50 |
| co-sberbank-capital-aliases | many-names | 40 | 13 | 50 | 50 |
| dense-putin | dense | 30 | 13 | 50 | 50 |
| dense-rotenberg | dense | 53 | 21 | 50 | 50 |
| dense-deripaska | dense | 19 | 17 | 50 | 50 |
| dense-tupikin | dense | 27 | 21 | 50 | 50 |
| dense-rosneft | dense | 16 | 13 | 50 | 50 |
| dense-huawei | dense | 13 | 11 | 50 | 50 |
| dense-stroygaz | dense | 8 | 21 | 50 | 50 |
| dense-negative-person | dense | 14 | 10 | 50 | 50 |
| long-name-string | long | 93 | 67 | 50 | 50 |
| long-address | long | 8 | 10 | 50 | 50 |
| degenerate-diacritics | long | 4 | 6 | 50 | 50 |
| degenerate-numeric-name | long | 1 | 0 | 50 | 0 |
| degenerate-repeated-words | long | 6 | 2 | 50 | 50 |

`names-ermakov-20` (22 unique parts, one fuzzy clause each) is the one query that got materially slower; `dense-stroygaz` is the other. Hamas and Hezbollah now return 36 and 12 candidates rather than a window padded with trigram matches; the numeric-only company name `12345` returns nothing instead of 50 trigram neighbours.

## Known limits

- Common single-token queries ("Ali", "Muhammad", "Maria") have thousands of exact matches for 50 slots; they all tie on score and `entity_id` order decides. Previously BM25 on alias count decided. Whether `entity_values_count` should break ties for `/match` is a separate decision.
- Edits at the first letter (wagner/vagner, gaddafi/qadhafi) are not retrieved unless a NAME symbol bridges them; two-edit first-letter changes need one. A second one-edit clause without the prefix constraint is the documented upgrade if this class matters.
- A merged query whose expansion is only a sub-sequence of a longer indexed name (`vladimirputin` against an entity that only has "Vladimir Vladimirovich Putin") is not reached; `name_joined` is exact.
- Two edits on a short part (chol/chun) and single-part Arabic edits are outside `AUTO` with a fixed first letter.
- The `names.ngrams` removal drops the reversed-string and split-token cases that trigrams happened to catch.

## Removed, and why

`representative_names` (clause count no longer depends on alias count), `SYMBOL_BOOSTS` (IDF already ranks org-class and location parts near zero), the `is_short` rule (short parts get exact and symbol channels; `AUTO` gives 0 edits under 3 chars anyway), `Name.consolidate_names` at query time (the one place query and index analysis differed), the `match names` clause and its ×3 boost (the alias penalty of #1252), the n-gram clauses and analyzer (#1254; trigrams at 70% have 0% overlap on john/jon and admit 20–1000× the documents), the phonetic channel (never the only bridge for a variant pair on the fixtures, unavailable for the scripts where a cross-script bridge is needed), `NAME_PHONETIC_FIELD` and `NAME_NGRAMS_FIELD`.

## Operational

`INDEX_VERSION` is bumped to `019`: the mapping changed, so deploying this triggers a full reindex of every dataset, and the old query code cannot serve from a 019 index nor the new code from an 018 one. `YENTE_MATCH_FUZZY` now toggles only the fuzzy sub-clause; exact, symbol and joined matching are always on, so merged names ("alexandervyacheslavovichzakharov") are retrieved regardless of the setting.

## Tests

`tests/test_queries.py` is new: one `dis_max` per unique part across seven transliterations, fuzzy sub-clause shape and absence when `MATCH_FUZZY` is off or the part is under three characters, symbol sub-clause presence and cap, joined clause contents, weak alias clause, the 100-part cap, and primary-before-alias ordering. `test_mappings.py` asserts `name_joined` is indexed (and `name_phonetic` is gone); `test_match_name_without_spaces` now asserts the merged name works with fuzzy on and off. `ruff`, `ruff format` and `mypy --strict` pass; the full suite passes except the four pre-existing `httpx_mock` fixture errors in `test_loader.py`/`test_updater.py` (pytest-httpx not installed in the local environment).


pudo:
Underlying prompt: 



I feel like the `entity_names` function and it's indexing counterpart have gone through a lot of evolution, and some of the contained decisions (representative_names, metaphone use) may not longer be ideal. I would like to come up with a simple alternative that makes ideal use of the available primitives (maybe (!) ngrams, symbols, name objects more generally) from first principles. Some thoughts on this:

* we want to retrieve based on all `matchable` names, which can be just one (most common) or many. We don't want to ever exceed to clause limit of ES cloud (4096?). 
* if a person has N aliases including the name part John, that usually doesn't mean they're a better (more relevant) "john" than a sparser record with just one "john" name part. this is true of both queries and results. the exception here is a person or organisation where the token is actually mentioned many times ("John Mc John"), but this seems exceedingly rare and maybe not a thing the candidate generation query needs to consider?
* the goal is to generate candidates for the symbol-based and fuzzy string matching process. a big ambition is the ability to retrieve candidates with spelling mistakes because then the fuzzy matcher can still make a positive decision. Term searches alone won't quite do this. 
* we want to match names across scripts and languages. transliteration of `comparable`, name symbols, metaphone and ES-side asciifolding are all theories about how we get this done, but some of them are duplicative, and metaphone is badly implemented - can we drop asciifolding and/or metaphone without loss of recall?  I am particularly questioning the continued use of asciifolding in this context.
* ngrams and fuzzy matching using a `match` query serve the same goal, do we ever need both?

Other failure modes:
* Multiple mis-spelled name parts (jon smyth ~ john smith)
* Merged names (alqaeda ~ al qaeda, vladimirputin ~ vladimir putin)

For context, see existing work in the name matching benchmarks, and these issues:
* https://github.com/opensanctions/yente/issues/1252
* https://github.com/opensanctions/yente/issues/1253
* https://github.com/opensanctions/yente/issues/1254 

The goal here is not to come up with a more complex, but with a really simple, explainable, error-reistant and reliable logic.

Other rules:
* Reindexing is cheap, we can always change the index format for free. 
* We can add stuff to rigour, but we want to keep the core of the indexing logic in yente.

Let's leave the weak alias query logic untouched, that's not a goal for this.

leonhandreke:
Cool! Some thoughts:


- Damn, the deletion thing is really smart, and I'm really surprised that Elastic has nothing built-in for it. I remember when I worked on n-grams the thing also suggested it, but I think I discarded it because intuitively, I was looking for something built into Elastic and nothing custom to maintain. But really, the amount of added code is really small and I can intuitively understand the function, so I think it's totally fine.
- With name parts, we are going per-token again, so omitting or adding a space somewhere is fatal again. I guess you can't have it all :(

PR #1258, opened 2026-09-13
Precompute deletion variants of name parts for fuzzy candidate retrieval

pudo:
Stacked on #1257 (base branch `feat/name-retrieval-redesign`); only the last three commits are this PR. Part of the fuzzy retrieval rework in #1254.

Replaces the query-time `fuzzy` clause of the `/match` name query with exact lookups against deletion variants that are precomputed at index time (the symmetric-delete scheme, as in SymSpell).

## The scheme

For every comparable name part, `name_part_variants(part)` returns the part itself plus every string obtained by deleting up to *k* of its characters, with *k* following the same length bands as Elasticsearch fuzziness `AUTO`: no deletions up to 2 characters, one at 3–5, two from 6. The indexer writes the variants of every indexed part into a new keyword field `name_part_variants`; the query builds the variants of every query part and sends them as one `terms` clause per part. Both sides use the same function, in `yente/data/util.py`.

Two strings within *k* Damerau-Levenshtein edits always share a variant: a substitution is bridged by deleting the differing character on both sides, an insertion or deletion by deleting it on the longer side, a transposition by deleting one of the swapped characters on both sides. So the `terms` clause admits everything the `fuzzy AUTO` clause admitted. Because deletion is position-blind, it also admits edits at the first letter (wagner/vagner, gaddafi/qadhafi, "Zlexander Sakharov"), which `prefix_length: 1` excluded. Sharing a *k*-deletion only bounds the plain Levenshtein distance by 2*k*, so the clause also admits some pairs up to twice the band apart, e.g. alexander/aleksandr at distance 3 through the shared variant `aleandr`, and a 5-character query (k=1) meets a 7-character indexed part (k=2) at distance 2. This is a strict superset of what fuzzy retrieved; the gate below measures how much wider it is and whether that costs recall.

Variant counts per part are `1 + n` at k=1 and `1 + n + n(n-1)/2` at k=2: 22 values for a 6-character part, 46 for 9, 79 for 12, 254 for 22. The largest single clause on the bench queries has 121 values, against an `index.max_terms_count` default of 65,536.

## What it replaces, and why

Prior art: `leonhandreke/fuzzy-deletions` (May 2026) built the same scheme with one-deletion variants and left a failing two-edit test to document the gap; this PR is that design with the deletion depth banded by part length, so the two-edit case is covered.

The `fuzzy` clause made Lucene compile a Levenshtein automaton per query part and intersect it with the term dictionary of every segment (1.24M distinct `name_parts` terms), which is where the 2–3 ms per part and 40–60 ms on 20-part queries went. It also had to be wrapped in `constant_score` in #1257, because a bare fuzzy clause sums the expansions a document carries and handed heavily aliased records an alias bonus. A `terms` query is one dictionary seek per value plus a postings union, and is constant-scored by ES, so the wrapper, `FUZZINESS`, `FUZZY_PREFIX_LENGTH`, `FUZZY_MAX_EXPANSIONS` and `FUZZY_MIN_LENGTH` are gone. The clause keeps the constant boost the fuzzy clause had (`VARIANTS_BOOST = 6.0`), so exact hits still outrank approximate ones exactly as in #1257; the rest of `names_query` is unchanged. `YENTE_MATCH_FUZZY` is removed: it was added in 2023 as a cost switch for the Levenshtein queries inside Elasticsearch, and the operation it guarded now costs about a millisecond, so there is nothing left to switch off. Setting the variable logs a warning and has no effect. Parts under three characters get no variants clause, since their only variant is the part itself, which the exact `term` already covers.

## Gate

Full local reindex of the `default` collection (4,049,380 documents, release `20260913185433-ggx`) into a 020 index: **13 min 00 s** wall, against 9 min 45 s for the 019 build of #1257. "PR #1257" columns are the stored results of #1257's gate on its 019 index (release `20260913125433-cac`); "#1257 code, same index" runs #1257's query code against this PR's 020 index, which still carries `name_parts`, so it is the like-for-like control. Fixtures from `contrib/validation_report/build/fixtures`, recall@50 = expected id present in a 50-candidate window, query built exactly as the router builds it.

### Recall@50

| fixture | n | main (018) | PR #1257 (019) | #1257 code on 020 | this PR (020) | median rank main / #1257 / this |
|---|---|---|---|---|---|---|
| positives_merged_aliases | 200 | 85.0% (170) | 99.5% (199) | 99.5% (199) | 99.5% (199) | 2 / 1 / 1 |
| positives_name_bench | 246 | 95.9% (236) | 97.6% (240) | 97.6% (240) | **98.0% (241)** | 1 / 1 / 1 |
| positives_un_treated | 732 | 96.6% (707) | 96.6% (707) | 96.6% (707) | 96.6% (707) | 1 / 1 / 1 |
| positives_un_untreated | 732 | 96.6% (707) | 96.6% (707) | 96.6% (707) | 96.6% (707) | 1 / 1 / 1 |
| positives_us_congress_untreated | 1321 | 98.6% (1302) | 98.7% (1304) | 98.7% (1304) | 98.7% (1304) | 1 / 1 / 1 |

No fixture loses a case. The one miss that changed, name-bench `عبد الفتاح السري` → `عبد الفتاح السيسي` [Q307871], was outside the window under #1257 (both indices) and is at rank 20 now. Arabic is not latinised, so the comparables are `السري` (5 characters, one deletion) and `السيسي` (6 characters, two deletions); they are two edits apart, outside fuzzy `AUTO`'s one-edit band for the 5-character query part, but share the variant `السي` because the budget is set per string.

Rank movement inside the window, this PR vs #1257 code on the same index (cases found by both):

| fixture | improved | same | worsened | left top-10 | entered top-10 |
|---|---|---|---|---|---|
| positives_merged_aliases | 0 | 199 | 0 | 0 | 0 |
| positives_name_bench | 2 | 211 | 27 | 4 | 1 |
| positives_un_treated | 3 | 703 | 1 | 0 | 2 |
| positives_un_untreated | 0 | 707 | 0 | 0 | 0 |
| positives_us_congress_untreated | 0 | 1301 | 3 | 0 | 0 |

The worsened ranks are the flat-tier effect: every variant hit scores the same 6.0 whether it is one edit away or three, so on a query whose exact matches do not fill the window, distance-3 neighbours tie with distance-1 ones and `entity_id` order decides. Most moves are one or two positions (Olaf Schulze 2→3, Serei Lavrov 1→2, Jeff Brezos 1→4); the largest are "Nikolai Patr ushev" 2→11, "Chevon Corporation" 4→13 and "Vladimir Pulin" 7→14. The gains are the classes this scheme adds: HOSEIN ASLIMI → Amirhossein Salimi 30→3, MOHAMMED MUAMMA RQADHAFI 39→2, BADULMALIK AL-HOUTHI 5→1, Lee Jae Hyung 12→3. All of these stay inside the window that the scorer then re-ranks.

### Latency (`contrib/perf_bench`, 100 queries, median of warm runs)

| stage | main p50 | PR #1257 p50 | this PR p50 | main p95 | PR #1257 p95 | this PR p95 | #1257 code on 020, p95 |
|---|---|---|---|---|---|---|---|
| search (ms) | 8.4 | 8.4 | **7.2** | 41.9 | 35.0 | **18.2** | 26.0 |
| ES took (ms) | 5.0 | 5.5 | **4.0** | 38.1 | 31.1 | **14.0** | 22.1 |
| score (ms) | 7.5 | 6.5 | 6.5 | 70.7 | 48.2 | 48.4 | 49.4 |

| query | took main | took PR #1257 | took this PR | took #1257 code on 020 |
|---|---|---|---|---|
| names-ermakov-20 | 28 | 58 | **16** | 64 |
| names-stroygaz-40 | 40 | 40 | **20** | 36 |
| dense-stroygaz | 8 | 21 | **8** | 20 |
| dense-rotenberg | 53 | 21 | **11** | 21 |
| names-kim-12 | 46 | 25 | **14** | 22 |
| long-name-string | 93 | 67 | **32** | 65 |
| dense-putin | 30 | 13 | 12 | 13 |
| names-zakharov-7 | 34 | 20 | **7** | 21 |
| names-putin-10 | 34 | 31 | **14** | 26 |
| co-sberbank-capital-aliases | 40 | 13 | 10 | 14 |
| sparse-ali | 7 | 4 | 5 | 5 |
| sparse-wagner | 5 | 3 | 2 | 4 |
| sparse-putin | 7 | 5 | 3 | 6 |
| sparse-hamas | 3 | 2 | 1 | 2 |
| sparse-hezbollah | 3 | 2 | 3 | 2 |
| sparse-rosneft | 3 | 3 | 1 | 3 |
| sparse-ivan-petrov | 6 | 5 | 3 | 7 |
| neg-john-smith | 5 | 3 | 3 | 3 |
| neg-mohammed-ali | 5 | 5 | 6 | 5 |
| pos-rotenberg | 10 | 11 | 7 | 10 |

The many-part queries that #1257 made slower (ermakov 28→58, dense-stroygaz 8→21) are now faster than on main; the search-stage p95 halves. Scoring is unchanged and is now the dominant stage on every query.

### Breadth (documents admitted by the whole candidate query, `_count`, same 020 index)

| query | #1257 code | this PR | ratio |
|---|---|---|---|
| names-ermakov-20 | 12,635 | 28,085 | 2.22× |
| names-kim-12 | 28,041 | 49,292 | 1.76× |
| long-name-string | 20,632 | 30,851 | 1.50× |
| dense-putin | 95,847 | 98,027 | 1.02× |
| dense-rotenberg | 91,910 | 96,791 | 1.05× |
| neg-john-smith | 15,104 | 16,880 | 1.12× |
| neg-mohammed-ali | 28,472 | 37,080 | 1.30× |
| sparse-ali | 16,066 | 24,693 | 1.54× |
| sparse-wagner | 851 | 1,507 | 1.77× |
| names-stroygaz-40 | 240,886 | 244,052 | 1.01× |

The wider admission comes from the short end of each band (6-character parts with two deletions leave 4-character keys; 3–4-character parts with one deletion leave 2–3-character keys). It costs nothing measurable in latency, since the postings union is cheap next to the dictionary walk it replaces, and does not displace any fixture case from the window.

### Index size

| | 018 (main) | 019 (#1257) | 020 (this PR) |
|---|---|---|---|
| store size, 4.05M docs | 2.5 GB | 2,645 MB | 4.0 GB (4,309 MB after merging to 24 segments) |

`_disk_usage` on the 020 index: `name_part_variants` 1.3 GB in total, of which 628 MB inverted index and 721 MB doc values; `name_parts` 60 MB (29 + 31), `name_joined` 88 MB (44 + 44), `name_symbols` 39 MB. Cardinality (`precision_threshold` 40,000): 33.7M distinct variants against 1.25M distinct `name_parts`, on the 1.72M documents that carry names. The inverted index is what the query uses; nothing sorts, aggregates or scripts on the field, so the mapping now sets `doc_values: false` on it (the 1.3 GB figure above was measured before that change). Expected size of the field after the change is the 628 MB inverted index alone, saving the 721 MB of doc values; not re-measured on a full reindex.

## Operational

`INDEX_VERSION` is bumped to `020`: the mapping gains a field and the query needs it, so deploying triggers a full reindex, and this code cannot serve from a 019 index. Index build is ~30% slower than 019 (variant generation is pure Python in `build_indexable_entity_doc`).

## Known limits

- **Flat scoring inside the fuzzy tier.** A `terms` hit scores the same constant at distance 1 and distance 3, where `fuzzy` graded by edit distance. Measured as the small within-window rank moves above; the candidates still reach the scorer.
- **Breadth on short parts.** 1.0–2.2× the documents of the fuzzy clause on the bench queries, from 3–6-character parts whose deletion keys are 2–4 characters long.
- **Over-budget admissions.** Pairs up to 2*k* apart can share a key; that is what brings alexander/aleksandr in without a symbol, and also what admits valery ↔ salter (both leave `aler`).
- **Very long parts** get `n(n-1)/2` variants each (254 values for 22 characters). Well under `max_terms_count`, and long parts are 0.2% of postings, so no cap is applied; one could fall back to k=1 above ~16 characters if index size becomes a concern.
- Reversed or rotated tokens (`rimaldiV nituP`) and pairs beyond the 2*k* bound remain out of reach, as under #1257.

## Tests

`tests/test_util.py`: variant counts per band, ≤2-character parts yield only themselves, the four edit types at any position including the first letter share a variant, thirty spelling-variant pairs (mohammed/muhammad, alexander/aleksandr, wagner/vagner, catherine/katherine, …) share a variant and unrelated pairs (john/smith) do not. `tests/test_queries.py`: the variants clause shape, its absence for parts under three characters, exact `term` still present. `tests/test_mappings.py`: variants indexed for a fixture entity, none for weak aliases, and a misspelled part reaches the indexed document through the field. `tests/test_match.py`: `test_fuzzy_names` and `test_match_name_without_spaces` unchanged and passing; new `test_fuzzy_names_first_letter` ("Zlexander Sakharov") retrieves Zakharov. `ruff`, `ruff format` and `mypy --strict` pass (the two pre-existing aiofiles/boto3 stub notes aside); the full suite passes except the four pre-existing `httpx_mock` fixture errors in `test_loader.py`/`test_updater.py`.


pudo:
Resolved the doc-values follow-up from the body in 5b97b2a: `name_part_variants` is now mapped with `doc_values: false` (only that field; `name_parts` and `name_joined` unchanged). Nothing in `yente/` or `contrib/` sorts, aggregates or scripts on the field, and `test_name_part_variants_mapping` asserts the mapping. On the gate's full `default` reindex the field's doc values measured 721 MB next to a 628 MB inverted index, so the expected saving is ~721 MB (index store roughly 4.3 → 3.6 GB); not re-measured with another reindex. No further `INDEX_VERSION` bump, 020 is new in this PR.

PR #1262, opened 2026-09-18
Fix error status codes

### LINKED ISSUE #1263: Statuses declared on YenteError subclasses are dead code
`yente/exc.py` binds the default at class-definition time:

```python
class YenteIndexError(YenteError):
    STATUS = 500
    def __init__(self, detail: str, status: int = STATUS, ...):
```

`status: int = STATUS` evaluates to `YenteIndexError.STATUS` (500) when the class body runs, so a subclass `STATUS` is never used: `IndexNotReadyError` answers 500 rather than 503, `YenteNotFoundError` 500 rather than 404. Relevant here because an index that isn't ready while shards are reallocating after a node leaves answers 500, which clients give up on. Defaulting `status` to `None` and reading `self.STATUS` fixes it.

baldurh:
Fixes #1263

## Answer errors with the status they mean

Three places reported a status that told the caller the wrong thing, and in each of them a search worth retrying read as one to give up on (400).

**1. A status declared on a `YenteError` subclass was never read.** `YenteIndexError.__init__` took `status: int = STATUS` — a default argument, bound to `YenteIndexError.STATUS` when the class was defined, so a subclass could declare whatever it liked and never be heard. `IndexNotReadyError` answered 503 in name and 500 in practice; `YenteNotFoundError` 404 in name and 500 in practice. The status is now read off the class when the error is raised; passing `status=` still overrides it.

**2. `search_phase_execution_exception` was reported as 400 unconditionally.** The index raises this both for a query it could not run and for a query it could not run on every shard, and only the status tells the two apart — 400 for a bad query, 429 for a full queue, 5xx for a lost shard. Reported as 400, all of them tell the client its request was malformed, so it gives up instead of retrying. The status the index gave is now the status the caller sees, in both providers.

**3. (debatable) `/readyz` answered "not found" for an index that is not ready.** `check_health` raised `YenteNotFoundError`; it backs `/readyz`, which documents 503, and a missing index there is the initial ingestion still running. It now raises `IndexNotReadyError`. `refresh()` keeps `YenteNotFoundError` — that runs in the indexer, where a missing index really is a mistake.


leonhandreke:
This look really great, thank you Baldur!

PR #1265, opened 2026-09-21
Disallow partial search results

### LINKED ISSUE #1260: With more than one scope configured, a lost shard silently drops results instead of erroring
Disclaimer: The issue raised here was identified by Claude and at the bottom is it's report

We’ve been seeing some failures in our daily OFAC list test where seemingly random names failed to return the expected results.

It seems the issue comes up when there is more than one index being used. In that case if a shard fails it does not raise an error but instead just returns whatever the remaining shards returned. The behavior can be changed using Elasticsearch's [`allow_partial_search_results`](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-search#operation-search-allow_partial_search_results) query parameter to force all shards to respond and otherwise return an error.

Here is our manifest:

```
catalogs:
  - url: "https://delivery.opensanctions.com/datasets/latest/index.json"
    auth_token: "$OPENSANCTIONS_DELIVERY_TOKEN"
    scope: default
    resource_name: entities.ftm.json
  - url: "https://delivery.opensanctions.com/datasets/latest/kyb/catalog.json"
    auth_token: "$OPENSANCTIONS_DELIVERY_TOKEN"
    scope: iso9362_bic
    resource_name: entities.ftm.json
```

This creates two indexes under a single alias. If the default index fails to respond only the results from iso9362_bic are returned—and for OFAC or other sanctions lookups—is a critical regulatory issue.

# Claude report

yente creates one index per **scope** — `manifest.py` sets `load` only on datasets named as scopes, and a collection scope like `default` becomes a single index however many source datasets it contains. Those indices are aliased together and searched as one.

How a shard failure surfaces depends on how many scopes there are, because Elasticsearch returns partial results only when at least one shard succeeds:

- **One scope** (the stock single-catalog config): one index, so a lost shard is *all* shards. The search fails outright and the caller sees an error. Visible, and correct.
- **Two or more scopes**: one index can go quiet while another answers. The response is a `200` that is missing an entire scope, with nothing to distinguish it from a complete answer.

A second scope most often comes from a second `catalogs:` entry, or from listing several `scopes:` under one catalog (as the commented-out example in `manifests/default.yml` shows). On our deployment it is a sanctions collection alongside a small reference dataset — the reference index answering is enough to make the response look fine, while the sanctions index is the one that went quiet. For screening, that reads as a name that is not on the list.

`successful >= 1` is not a usable check for the same reason.

### The mechanism

`SearchProvider.search()` does not set `allow_partial_search_results`, so Elasticsearch's default of `true` applies and an unreachable shard is simply left out of the results.

The `_shards` counts are the only in-band record that a search ran on less than the whole index, and `search()` returns `response.body` without reading them, so nothing downstream can tell a whole answer from a partial one.

`yente/provider/elastic.py` (same in `opensearch.py`):

```python
response = await self.client().search(index=index, query=query, ...)
return cast(dict[str, Any], response.body)
```

### Why a replica doesn't cover it

Normally it would. Elasticsearch queries one copy per shard, and if that copy fails it advances to the next one in the shard iterator; a shard only counts as failed once every copy is exhausted.

That retry appears not to be available on the `dfs_query_then_fetch` path, which `search_entities` uses via `rank_precise=True`. The DFS phase opens a reader context on one specific copy to collect term statistics, and the query phase sends only that context ID back to the same copy. If the node goes away in between, there is no context to retry with on another copy. The error captured below names `indices:data/read/search[phase/query/id]` — the context-bound query action, not the self-contained one used by plain `query_then_fetch`.

This is inferred from the action name and the observed behaviour rather than confirmed against the ES source, so treat it as the likely mechanism. Happy to be corrected.

### Reproduction

Three-node Elasticsearch 9.5.1, one shard per index with a copy on every node, two indices behind one alias (a clean search reports `{"total": 2, "successful": 2, "failed": 0}`). A node was killed while searches were in flight, against 150k documents with queries taking 0.2–0.3s and `search_type=dfs_query_then_fetch` as yente issues them:

| | `allow_partial_search_results` unset | set to `False` |
|---|---|---|
| searches | 884 | 906 |
| answered `200` without the entity that was on the list | 9 (1.0%) | 0 |
| errors | 6 (0.7%) | 16 (1.8%) |

```
HTTP 502  error.type = search_phase_execution_exception
  reason      : "Partial shards failure"
  root cause  : node_disconnected_exception
                "[es0][...][indices:data/read/search[phase/query/id]] disconnected"
```

An earlier attempt over a one-document index found nothing across 95,894 searches — it answers in microseconds and leaves no window to be interrupted in. Realistic document counts and query times are needed to reproduce.

### Suggested fix

1. Pass `allow_partial_search_results=False` on search. Elasticsearch still tries every available copy before failing, so a healthy cluster is unaffected.
2. Log the `_shards` counts when `failed > 0` or `timed_out`.
3. Make the resulting error retryable. `search_phase_execution_exception` is currently reported as `400` unconditionally:

   ```python
   if ae.error == "search_phase_execution_exception":
       raise YenteIndexError(f"Search error: {ae!s}", status=400) from ae
   ```

   Elasticsearch raises the same exception type for a query it cannot run (4xx) and one it could not run on every shard (5xx). As a `400` a lost shard reads as "bad query" and clients won't retry it. Without this, change 1 trades a silent under-match for a non-retried hard failure. An HTTP-level retry recovers cleanly, since it starts a fresh DFS phase on a healthy copy.

   `429` wants carrying through as well, not just 5xx — Elasticsearch answers `429` for `search_phase_execution_exception` wrapping `es_rejected_execution_exception` (search thread pool queue full), and disallowing partial results makes that more likely, since one rejected shard then fails the whole search.

Changes 1 and 2 are worth doing regardless of scope count: a single-scope deployment still can't tell from the response whether a search timed out or skipped shards.

leonhandreke:
Fixes https://github.com/opensanctions/yente/issues/1260

With several scope indices behind one alias, a lost shard made Elasticsearch answer 200 with a whole scope missing from the hits. For a screening lookup, that reads as a name that is not on the list.

Both providers now pass `allow_partial_search_results=False` on the search call. A lost shard fails the search instead. A healthy cluster is unaffected, because Elasticsearch still tries every copy of a shard before it gives up.

The failed search reaches the client with the status the index gave, so it is retryable. https://github.com/opensanctions/yente/pull/1262 made that change.

The issue also suggests to log the `_shards` counts when a response has failed shards or timed out. This PR does not do that. With partial results disallowed, a shard failure is an error and never a 200 response. Yente sets no search timeout, so `timed_out` is never true.

🤖 Generated with Claude Code using Claude Fable 5.1


baldurh:
> A healthy cluster is unaffected, because Elasticsearch still tries every copy of a shard before it gives up.

I don't believe this holds since `search_entities` uses `dfs_query_then_fetch` which  appears to remove the fallback. The query phase is bound to the copy that served the DFS phase, so a lost copy marks the shard failed without trying another. That's independent of this change really. `allow_partial_search_results=False` only decides whether the failed shard is dropped silently or fails the search.

So basically this will result in more errors, which is still better than silent failures.

But it's worth noting that because of `dfs_query_then_fetch` there has never been a multi-copy retry mechanic, even when there's only a single index (it was added to handle scoring when using more than one index https://discuss.elastic.co/t/querying-an-alias-throws-off-scoring-completely/351423/4). Perhaps it can be used only when there is more than one index? Not sure if yente knows that at runtime.

Actually, it would probably be best if both of these (`dfs_query_then_fetch` + `allow_partial_search_results=False`) were only used when there is more than one index.

leonhandreke:
Thanks for your feedback!

On `allow_partial_search_results=False` for single-shard searches - IIUC, this shouldn't make a difference, so we can just always keep it?

On `dfs_query_then_fetch`: yeah, ideally we'd disable it for single-shard queries, i.e. `settings.INDEX_SHARDS > 1` or more than one dataset configured in the manifest. However, I don't consider myself very knowledgeable about Elastic, I wonder if there is an edgecase where through some administrative action or setting the number of shards gets overridden? Therefore, I'd intuitively be a bit scared to special case this here.

Further, I let the LLM (in this case `gpt-5.6-sol`) loose on it and it found that Elastic already performs this optimization internally: https://github.com/elastic/elasticsearch/blob/v9.4.2/server/src/main/java/org/elasticsearch/action/search/TransportSearchAction.java#L789-L807. I hope this is true, since I don't have the expertise to reasonably verify it, but it sounds plausible?

baldurh:
yeah seems you're right :)

leonhandreke:
Cool, thanks for flagging this and prompting a deeper (LLM-based, but still) investigation :)

